PR #2, opened 2015-07-24
Static map in all the places

JoeShep:
Tweaking the static map to accept data from the view helper instead of a yaml file.


jcreamer898:
Noice.

PR #7, opened 2015-08-03
Fix the race condition with brightcove... finally.

jcreamer898:
Verify by going to `/belize`, pressing play.

You'll see the video load, and then the close button will fade in.

Click the close button and make sure the video actually stops.


edubkendo:
:+1:

PR #17, opened 2015-08-11
2 quick fixes

jcreamer898:
Check for love letter read more, simple small thing to add class rather than create new element.


edubkendo:
:+1:

PR #18, opened 2015-08-11
Fixing spacing for love letter author image.

runako:


runako:
:+1: from @jcreamer898 on Slack.

PR #20, opened 2015-08-12
Limit static map height

JoeShep:
Addresses the ever-expanding map size as the screen widens. Now behaves similarly to hotels, where the container height stays static and the background image zooms to fit the container.
Also adds the 'open map' button and moves the logic for triggering a new interactive map said button into the static map


JoeShep:
@jcreamer898 :+1:

PR #23, opened 2015-08-14
Refactor Survival Guide to display new icon SVGs

JoeShep:
Updates the logic to use icons stored locally as data URIs. 
Note that there are not yet icons for every possible Survival Guide item. As is, some items will display without an icon. The most likely items (visas, best time to go, money and costs, health)  all have icons. But if we want to wait until more icons are produced, then I will resist the urge to merge.

Goes with this DN PR: https://github.com/lonelyplanet/destinations-next/pull/297


src/components/survival_guide/_survival_guide.scss
```diff
@@ -62,6 +62,13 @@ $mobile-breakpoint: 400;
   }
 }
 
+.survival_guide__list__item__icon {
```
jcreamer898: We probably should refactor this file a bit. The BEM is a little odd. Should be...

```sass
.survival-guide {
  &__list {
  }

  &__link {

  }

  &__item {
    &--even {

    }
  }
}
```

We probably need to do this across the board and try and wrap our heads around BEM  better.

JoeShep:
@jcreamer898 Check the BEM-iness. Also cleaned out the styling for even and odd # layouts, since we now always show 4 or nothing.

jcreamer898:
:shipit:

PR #34, opened 2015-08-26
Adding top neighborhoods.

runako:


src/components/masthead/masthead.hbs
```diff
@@ -33,11 +33,18 @@ found at https://accounts.brightcove.com/en/terms-and-conditions/.
...
-          </a>
+          {{#if top_places.neighborhoods}}
+            <a class="masthead__button js-top-places" >
+              Top Neighborhoods in {{place.name}} <i class="icon-chevron-down"></i>
+            </a>
+          {{else}}
+            <a class="masthead__button js-top-places" >
+              Top Places in {{place.name}} <i class="icon-chevron-down"></i>
+            </a>
+          {{/if}}
```
thomasthesecond: Instead of duplicating the markup, could you do something like so?

```handlebars
<a class="masthead__button js-top-places">
  Top {{#if top_places.neighborhoods}}Neighborhoods{{else}}Places{{/if}}
  in {{place.name}} <i class="icon-chevron-down"></i>
</a>
```

thomasthesecond:
:+1:

PR #37, opened 2015-08-28
Don't display food and drink image if not present

mkunkel:
We will be supplying these from ES. If no image exists, we should not show one.

### DIFF
--- a/src/components/food_and_drink/_food_and_drink.scss
+++ b/src/components/food_and_drink/_food_and_drink.scss
@@ -32,13 +32,6 @@ $fad-color: #b2b4b8;
     width: (100%/3);
 
     z-index: -1;
-
-    background: right center url(https://assets.staticlp.com/destinations-next/images/food-and-drink-default@1x.jpg) no-repeat;
-    background-size: contain;
-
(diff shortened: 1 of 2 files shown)


src/components/food_and_drink/_food_and_drink.scss
```diff
@@ -33,9 +33,6 @@ $fad-color: #b2b4b8;
 
     z-index: -1;
 
-    background: right center url(https://assets.staticlp.com/destinations-next/images/food-and-drink-default@1x.jpg) no-repeat;
-    background-size: contain;
-
     @media (-webkit-min-device-pixel-ratio: 1.3), (-o-min-device-pixel-ratio: 2.6 / 2), (min--moz-device-pixel-ratio: 1.3), (min-device-pixel-ratio: 1.3), (min-resolution: 1.3dppx) {
       background-image: url(https://assets.staticlp.com/destinations-next/images/food-and-drink-default@2x.jpg);
     }
```
thomasthesecond: Should we remove the 2x image as well?

mkunkel: Good catch.

thomasthesecond:
:+1:

PR #38, opened 2015-08-31
Design Tweaks: Round 1

thomasthesecond:
This is a first pass and I still have more things to update. My goal is to break up the PRs a bit so there aren't so many commits/changes. I could've done a better job with this one, though.


JoeShep:
Only slightly shorter than Moby Dick, but much more entertaining. :+1:

PR #39, opened 2015-08-31
Add new slideshow component

jcreamer898:
Creates a new slideshow component that can be re-used throughout the app.
Adds images to POIs in the detail tab and places in the about tab.


JoeShep:
This looks great! But highlights the need to deal with image 404s. The gray blocks gotta go. But I'm cool with giving this a :+1: and adding the missing image thing as a big report

PR #40, opened 2015-08-31
Updating sub-nav margin

thomasthesecond:
Removing margin at 480 breakpoint so that sub-nav border width matches that of the segment border width between 480 and 720.


jcreamer898:
:shipit:

PR #41, opened 2015-08-31
Updating food-and-drink component

thomasthesecond:
- Modifying font sizes and whitespace to match design comp
- Uncommenting hover transition and adding backface-visibility property to fix Chrome bug
- Adding adjacent selector for image
- Nesting MQs
- Adding inline style for food-and-drink__image background-image (although, not sure if this is how the image will be added)


jcreamer898:
Looks good dude.

PR #42, opened 2015-08-31
Updating interests component

thomasthesecond:
- Cleaning up a bit
- Updating font size, padding to match comp
- Changing `h1` to `h2` in hbs template
- Updating `$desktop-size` and `$fluid-max` variables
- Updating max-width breakpoint for `generate-dynamic-media-queries` mixin
- Wrapping hover transition with MQ for larger screens


jcreamer898:
:shipit:

PR #44, opened 2015-08-31
Interactive map POI items interactions update

JoeShep:
Updates the connected behavior of the map POI labels and their corresponding sidebar items.
-- Hovering on an item in the sidebar adds a background color to it
-- Clicking on an item in the sidebar opens the POI label in the map
-- Hovering a POI marker on the map highlights the corresponding item in the sidebar
-- Clicking a POI marker on the map loads the description in the sidebar
-- Map POI pop-up are moved to the right of the marker, instead of directly over the marker, for a less disorienting user experience

screen shot 2015-08-31 at 5 01 18 pm

### DIFF
--- a/src/components/map/actions.js
+++ b/src/components/map/actions.js
@@ -35,13 +35,13 @@ let MapActions = {
     Arkham.trigger("poi.closed");
   },
 
+  // looks like this wasn't being used?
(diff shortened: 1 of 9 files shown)


jcreamer898:
Love it.

PR #46, opened 2015-09-01
Text and link colors for breadcrumb

mkunkel:
Ensure that breadcrumb links and text (commas) are white for mastheads with images, and black (#333) for mastheads with no images.


jcreamer898:
:shipit:

PR #48, opened 2015-09-01
Map sidebar updates

jcreamer898:
Add buttons for detail and about view. 
Change the missing images. 
Extract topic images out of search items


sass/elements/_topics.scss
```diff
@@ -0,0 +1,101 @@
...
+
+    width: 8rem;
+    height: $height;
+
+    background: #ccc;
+
+    background:{
+      size: cover;
+      repeat: no-repeat;
+    }
```
thomasthesecond: With the `background`, you could probably move color here like so:
```scss
background: {
  color: #ccc;
  size: cover;
  repeat: no-repeat;
}
```

src/components/map/actions.js
```diff
@@ -12,12 +12,24 @@ let MapActions = {
   },
 
   gotoPlace: (data) => {
-    let place = data.place;
+    let place = data.place,
+        url = "/" + place + "/map.json";
+    
     Arkham.trigger("place.fetching", data);
+
+    // TODO: JC, maybe this is cool, maybe not?
```
thomasthesecond: Does this `TODO` need to be addressed?

src/components/map/views/item.jsx
```diff
@@ -20,7 +22,13 @@ export default class ItemView extends React.Component {
       }
     }
     if (item.geo.properties.image) {
-      imageSrc = "http://images-resrc.staticlp.com/S=H150/" + item.geo.properties.image;
+      let imgSrc = "http://images-resrc.staticlp.com/S=H150/" + item.geo.properties.image;
+      img = <img src={imgSrc} />
+    }
+    else {
+      // TODO: This will have to change when topics are correct
```
thomasthesecond: `TODO`

PR #49, opened 2015-09-02
Updating survival-guide component

thomasthesecond:
- Adding hover, focus states to link and giving style to icon and title
- Changing `h1` to `h2`
- Removing `survival-guide__inner`; there are no styles being applied to it
- Refactoring naming to remove unnecessary `__list` and `__list__item` prefixes


JoeShep:
:ship:

PR #50, opened 2015-09-03
Updating hotels (search) component

thomasthesecond:
This PR also includes updates to some mixins.

### DIFF
--- a/sass/_base_mixins.scss
+++ b/sass/_base_mixins.scss
@@ -43,8 +43,8 @@
     border-bottom: .1rem solid $color-border;
 
     @media (min-width: $min-720) {
-      margin-left: ($gutter*1);
-      margin-right: ($gutter*1);
+      margin-left: ($gutter);
+      margin-right: ($gutter);
 
       bottom: -$gutter*3 + $spacing;
     }
--- a/sass/_base_mixins.scss
+++ b/sass/_base_mixins.scss
@@ -1,14 +1,25 @@
 /// Use to separate components on the page.
-/// ### Modifiers 
+/// ### Modifiers
 /// * `--grey` modifier to change the background color
-/// 
+///
 /// @example css
-/// .segment { 
+/// .segment {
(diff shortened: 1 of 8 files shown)


sass/_base_mixins.scss
```diff
@@ -23,40 +34,36 @@
...
 
     max-width: $max-width;
     height: 0;
 
     border-bottom: .1rem solid $color-border;
 
-    @media (min-width: $min-720){
-      margin: 0 ($gutter*1);
+    @media (min-width: $min-720) {
+      margin-left: ($gutter*1);
```
jcreamer898: I don't think this means what you think it means... $gutter * 1 === $gutter?

jcreamer898: aka 30 * 1 === 30.

src/components/hotels/_hotels.scss
```diff
@@ -1,221 +1,176 @@
 @import "../../../sass/webpack_deps";
 @import "../../../sass/datepicker";
 
-$small-breakpoint: 600;
-$mobile-breakpoint: 717;
-$wide-breakpoint: 1024;
-$form-height: 4.2rem;
+$hotels-breakpoint-small-max: em(729px);
```
jcreamer898: Is there no way to use the `$min-xxx` breakpoints from the `_variables`?

jcreamer898:
Left a few comments, but otherwise... :shipit:

sass/_base_mixins.scss
thomasthesecond: Good catch. I think some of this was pre-existing but you're right, should just be `$gutter`.

src/components/hotels/_hotels.scss
thomasthesecond: I can see if any of those breakpoints match up.

jcreamer898:
Looks good! :+1:

PR #55, opened 2015-09-03
Updating guidebook component

thomasthesecond:
- Style updates to match design spec
- Updating markup
    - `article` to `div`
    - Renaming `guidebook__title-wrap` to `guidebook__header` and using `header` element
    - `h1` to `h2`, `h2` to `p`
- Removing some nesting in SCSS to improve scanability and searchability
- Removing unused variables and prefixing variables with `guidebook`
- Replacing `respond-to` mixin with plain CSS `@media`
- Utilizing `max-width` MQs to avoid property duplication/overriding
- Some clean up/formatting updates

### DIFF
--- a/src/components/guidebook/_guidebook.scss
+++ b/src/components/guidebook/_guidebook.scss
@@ -1,13 +1,10 @@
 @import "../../../sass/webpack_deps";
 
-$guidebook-breakpoint-large-max: 1023px;
-$guidebook-breakpoint-xlarge-min: 1024px;
-
 .guidebook {
   @include clearfix;
   margin: 0 auto $gutter;
 
-  @media (max-width: $guidebook-breakpoint-large-max) {
+  @media (max-width: $max-1080) {
     text-align: center;
   }
 }
@@ -16,7 +13,7 @@ $guidebook-breakpoint-xlarge-min: 1024px;
   position: relative;
   z-index: 5;
 
-  @media (min-width: $guidebook-breakpoint-xlarge-min) {
+  @media (min-width: $min-1080) {
     float: right;
     width: 35%;
     margin-right: 3%;
@@ -30,7 +27,7 @@ $guidebook-breakpoint-xlarge-min: 1024px;
   font-weight: $font-weight-medium;
   line-height: (32 / 18);
 
-  @media (min-width: $guidebook-breakpoint-xlarge-min) {
+  @media (min-width: $min-1080) {
     width: 100%;
     max-width: 400px;
     margin-left: 0;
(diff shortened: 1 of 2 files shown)


src/components/guidebook/_guidebook.scss
```diff
@@ -1,128 +1,112 @@
...
-    text-align: left;
-  }
 
-  &:after {
-    display: block;
-    visibility: hidden;
-    clear: both;
-    height: 0;
-    content: "";
+  @media (max-width: $guidebook-breakpoint-large-max) {
```
jcreamer898: `$min-1080`?

src/components/guidebook/_guidebook.scss
```diff
@@ -158,37 +138,34 @@ $wide-breakpoint: 1024;
...
-
-    @media (min-width: $min-560) {
-      margin-left: (-25rem / 2);
-    }
+.guidebook__picture--primary {
+  position: absolute;
+  z-index: 1;
+  top: 0;
+  left: 50%;
+  margin-left: (-20rem / 2);
```
jcreamer898: We have a nice `$gutter` variable you could use in these cases.

src/components/guidebook/guidebook.hbs
```diff
@@ -1,72 +1,67 @@
 {{#if books }}
-  <article id="guidebook" class="guidebook">
-    <div class="guidebook__title-wrap">
+  <div class="guidebook" id="guidebook">
```
jcreamer898: Any reason for changing this away from an article? Most of our other components are defined at the high level with article tags. Just curious.

thomasthesecond: Thanks for talking through this. I went and re-read the [spec](http://www.w3.org/TR/html5/sections.html#the-article-element). It says:

> The article element represents a complete, or self-contained, composition in a document, page, application, or site and that is, in principle, independently distributable or reusable, e.g. in syndication. This could be a forum post, a magazine or newspaper article, a blog entry, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.

After reading that, it makes sense to keep this an `article` since each component is an "independent item of content". I like the idea of using custom components or a `component` classname and maybe that's something we can look at in the future.

src/components/guidebook/_guidebook.scss
thomasthesecond: These margin numbers were pre-existing. It seems they're not using the `$gutter` because the margin is being used to position the guide image outside of the container and so they need to be specific numbers.

jcreamer898:
Nice

PR #57, opened 2015-09-04
Don't render a set if there are no items

jcreamer898:
Yeah, so whatever...


JoeShep:
http://d2f8dzk2mhcqts.cloudfront.net/0772_PEW_Roundup/09_Squirrel.jpg

PR #60, opened 2015-09-08
Fix issue with no masthead slides

jcreamer898:
Mexico was blowing up because it lacked a masthead. This should fix that.


edubkendo:
:+1:

PR #65, opened 2015-09-11
Map font tweaks and formatting

JoeShep:
Adding in the new fonts necessitated some resizing and spacing to bring the map look inline with the designs. Also added some minor formatting to clean up the "about" panel content


jcreamer898:
:shipit:

PR #68, opened 2015-09-16
Reference the s3 fonts

jcreamer898:
No reason to keep copying the fonts every deploy.

### DIFF
--- a/src/core/logger.js
+++ b/src/core/logger.js
@@ -12,7 +12,7 @@ export default class Logger {
    */
   error(err) {
     console.error(err);
-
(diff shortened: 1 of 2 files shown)


src/core/logger.js
```diff
@@ -12,7 +12,7 @@ export default class Logger {
    */
   error(err) {
     console.error(err);
-
+
```
thomasthesecond: Is this spaced needed?

jcreamer898: Who knows... lol

PR #69, opened 2015-09-17
Updating spacing in masthead

thomasthesecond:
- Fixing spacing when no strapline is visible
- Increasing bottom margin on breadcrumb

### DIFF
--- a/src/components/masthead/masthead_component.js
+++ b/src/components/masthead/masthead_component.js
@@ -23,6 +23,14 @@ export default class MastheadComponent extends Component {
 
     this.listenTo(this.slideshow, "image.changed", this.updateStrapline);
 
+    $.each(this.$straplines, function(index, strapline) {
+      if (!$(strapline).html()) {
+        $(strapline)
+          .parent()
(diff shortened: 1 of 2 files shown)


src/components/masthead/masthead_component.js
```diff
@@ -23,6 +23,14 @@ export default class MastheadComponent extends Component {
 
     this.listenTo(this.slideshow, "image.changed", this.updateStrapline);
 
+    $.each(this.$straplines, function(index, strapline) {
```
jcreamer898: I think this could be accomplished in the markup and might be slightly more effective there...

```html
<div class="masthead__straplines {{#unless strapline}}has-empty{{/unless}}">
```

Try this and see if it works.

thomasthesecond: It does, kinda, but it seems to always add the `has-empty` class, even when there are no empty straplines.

Try `/asia` (no empty) and `/spain/barcelona` (empty).

jcreamer898:
Just use the JS solution and :shipit:

PR #72, opened 2015-09-21
Design updates for hotels component

thomasthesecond:
- Changing 'stay' to 'sleep'
- Disabling outline on focus state
- Updating datepicker font sizes


edubkendo:
:+1:

PR #79, opened 2015-09-22
Updating segment and container mixins

thomasthesecond:
- Changing `padding` to `margin` on container so that padding does not subtract from actual width; this allows the the container to reach the max 1290 like it should, rather than 1170 (1290 - 120px padding)
- Changing media queries on segment to match those of container; removing 1170px max-width


JoeShep:
:+1:

PR #86, opened 2015-09-23
Center the map open btn more better

JoeShep:
Sets a specific btn width that can be divided by 2 to center properly.

### DIFF
--- a/src/components/map_static/_map_static.scss
+++ b/src/components/map_static/_map_static.scss
@@ -30,11 +30,14 @@
   }
 
   &__btn {
+    $btnWidth: 268px;
+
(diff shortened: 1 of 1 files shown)


edubkendo:
:+1:

thomasthesecond:
Here's another clever way to center something horizontally (or vertically). If you've got something `position: absolute` and it has a `width` (or `height`), you can add this:

```css
margin-left: auto;
margin-right: auto;
left: 0;
right: 0;
```

and to center vertically, change `left` and `right` to `top` and `bottom`.

You don't have to change what you have, I just wanted to share this.

PR #88, opened 2015-09-23
Using `$color-blue` variable where it was not

thomasthesecond:


src/components/navigation/_navigation.scss
```diff
@@ -36,7 +36,7 @@ $medium-breakpoint: 768;
 
     &--active a,
     &:hover a {
-      //color: #287bbb;
+      //color: $color-blue;
     }
```
jcreamer898: These are both commented out?

jcreamer898:
One small question, then good to go.

src/components/navigation/_navigation.scss
thomasthesecond: Just the one line, but I can probably remove it altogether.

PR #89, opened 2015-09-25
Fixing button overlap for short screens

thomasthesecond:
- Adding `max-height` media query to change play button appearance and to adjust title font size
- Moving play button in DOM
- Changing some `px` values to `rem` for consistency


jcreamer898:
:shipit:

PR #95, opened 2015-09-28
Fixing layout issues with map sidebar

thomasthesecond:
- "Close map and explore" shouldn't cover part of the text
- The tabs don't look well in Chrome


edubkendo:
thumbs_up5

PR #104, opened 2015-09-29
Add test coverage, test files, etc

jcreamer898:
Using istanbul loader to generate test coverage.
Added a bunch of empty specs that need to be filled in.


edubkendo:
clap
haruhithumbsup

PR #109, opened 2015-10-01
Remove /map from url or add it if needed

jcreamer898:
This will add or remove `/map` from the url when the map is opened and closed.


edubkendo:
this seems like a good candidate for a test

jcreamer898:
@edubkendo I like where your head is at.

edubkendo:
hooray
goodjob

PR #115, opened 2015-10-02
Working on update docs, and a few small rizzo enhancements

jcreamer898:
Customizes are docs output a bit. 
Also adds better docs to Component. 
Move component to a class.
Improve the ComponentRegistry.


edubkendo:
thumbs_up5

PR #120, opened 2015-10-06
Removing outline from guidebook pictures

thomasthesecond:
The markup has a `tabindex` of `-1` so the photos cannot be tabbed to because we want the button to have the focus, therefore removing the outline is ok in this case.


edubkendo:
haruhithumbsup

PR #121, opened 2015-10-06
Updating tabs, articles

thomasthesecond:
- Updating styles for small screen; using flexbox
- Making tab font size smaller
- Cleaning up a bit


jcreamer898:
Looks good.

PR #125, opened 2015-10-09
Updating guidebook sub header style

thomasthesecond:


edubkendo:
isaac-and-miria-thumbsup

PR #128, opened 2015-10-09
Wrap map close btn in clickable div

JoeShep:
This moves the 'close map' click event to a wrapper around the btn so the map closes when user clicks anywhere above or below the btn.


thomasthesecond:
:+1:

PR #129, opened 2015-10-09
Updating spacing and adding a min-height to maintain alignment

thomasthesecond:
Also fixing the jitter on hover with the number


thomasthesecond:
Forgot I need to make some updates to mobile first.

edubkendo:
isaac-and-miria-thumbsup

PR #130, opened 2015-10-09
Fix tours layout

JoeShep:
Fixes layout when there is an uneven # of tours. Now they stay to the left at all times.


edubkendo:
thumbsup1

PR #132, opened 2015-10-09
Make map sidebar render tab for other topics when no experiences

JoeShep:
Previously, when a place had no experiences, no tab was rendered in the map sidebar, which also meant no dropdown menu would render even if other things to do were present. A tab will now always be created as long as a topic with POIs exists.
Also adds highlighted styling for the first item in the drop down when it first renders.

screen shot 2015-10-09 at 4 34 32 pm


edubkendo:
thumbsup3

PR #137, opened 2015-10-12
Update sub-nav to fix overflow, visible scroll bar

thomasthesecond:
Adding an absolute height to the container and making the the list height taller for small screens to push the horizontal scroll bar off the screen.


edubkendo:
isaac-and-miria-thumbsup

PR #138, opened 2015-10-12
Fix accent image clipping

thomasthesecond:
- Using `img` element for image instead of a `div` with a background image; this allows the image to flow and size more naturally
- Cleaning up CSS


edubkendo:
goodjob

PR #140, opened 2015-10-13
Fix issue with clamping in the tours component

jcreamer898:


edubkendo:
dont_want_to_see
thumbsup1

thomasthesecond:
@jcreamer898 What was the issue here? I'm starting to see text overlapping and I'm thinking this is related. If I change the default back to `18`, I don't notice overlapping.

screen shot 2015-10-19 at 3 23 38 pm

PR #143, opened 2015-10-13
Make map item images background-image

JoeShep:
When images were set as tags and positioned absolutely, some were off-center or had important parts cut off. Setting them as background images allows for a much higher percentage of imgs to be presented correctly.
Before:
screen shot 2015-10-13 at 3 13 08 pm

After:
screen shot 2015-10-13 at 3 13 16 pm

### DIFF
--- a/src/components/map/sidebar/item.scss
+++ b/src/components/map/sidebar/item.scss
@@ -85,6 +85,7 @@
 
   &__pic {
     background-size: cover;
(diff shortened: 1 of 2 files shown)


thomasthesecond:
:+1:

thomasthesecond:
On second thought, you might want to add `background-repeat: no-repeat;` to `pic` as well.

PR #144, opened 2015-10-13
Use instanceof for component check. Add lcov for Karma

jcreamer898:
The `Object.getPrototypeOf` wasn't working in IE. This is a better check for inheritance.
Also adding in `lcov` output for Karma coverage reporting.


JoeShep:
:+1: in my expert opinion

PR #145, opened 2015-10-14
Update things to do component

thomasthesecond:
- Using `js-` classes
- Adding conditional in hbs template to show "See more" link


jcreamer898:
Seems legit.

PR #146, opened 2015-10-14
Update food and drink layout

thomasthesecond:
Previously, the food and drink component used CSS columns for layout. There were a couple of issues:

1. No support for IE 9. While there has been discussion about dropping support for this browser, it has not been made official.
2. The current implementation assumed that there would always be 10 items. There were some fragile `nth-child` conditions in the code that broke the layout when there weren't 10 items. For example, /ireland/northern-ireland/ballycastle has 5 items and with the `nth-last-child(-n+5)` condition, no items would appear on screens less than 960px. This has been fixed by adding an index to the markup and then with a Sass loop, we can explicitly hide the last five items, regardless of the number of total items. Also, the `nth-child(6)` condition was being used to add a border, but it also breaks when there aren't 6 items.

The solution was to replace CSS columns with floats. This does change how the items are ordered (left to right instead of top to bottom), but this solution ensures compatibility with older browers and makes our `nth-child` condtions less fragile.

Changing the layout to floats also allowed for removal of a couple of hacks to fix a bug with transitioned elements inside of column. Those were the `backface-visibility` and `transform` properties on the anchor element.

Lastly, a `z-index` of -1 was added back to the image element just in case the image and the heading should ever overlap, the image will not cover up the heading.

screen shot 2015-10-14 at 12 09 09 pm

screen shot 2015-10-14 at 12 09 31 pm

### DIFF
--- a/src/components/food_and_drink/_food_and_drink.scss
+++ b/src/components/food_and_drink/_food_and_drink.scss
@@ -28,6 +28,7 @@ $item-border-color: #f1f4fb;
       width: $fad-column-width;
       margin-top: auto;
       margin-bottom: auto;
+      z-index: -1;
     }
   }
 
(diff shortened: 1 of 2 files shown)


jcreamer898:
Awesome man. Great job on the description here too!

PR #147, opened 2015-10-14
Drop flexbox from tours layout

JoeShep:
To fix an earlier layout bug in the tours component Flexbox was added to make the items justify left when there was an uneven number of tour items. But we discovered that the layout broke in Safari, even though Flexbox is supported. Quickest fix was to drop flexbox and use good old fashioned floats.


thomasthesecond:
:+1:

PR #151, opened 2015-10-15
Make notification badge reusable

thomasthesecond:
This update removes some context specific styles (i.e., positioning) from the component's base styles and moves them to a modifier. This allows the badge to be reused in different contexts and doesn't require overriding certain properties.


jcreamer898:
Dude, love it man.

PR #153, opened 2015-10-15
Unstick sub-nav when it hits the bottom of the content area

thomasthesecond:
This update "unsticks" the sub-nav when it hits the bottom of the content area and prevents it from covering the footer. The issue mostly occurs on mobile-sized screens.


thomasthesecond:
https://github.com/lonelyplanet/rizzo-next/blob/tc-sub-nav-scroll-stop/src/components/sub_nav/index.js#L12-L14

The delay is needed because on initial load, the height is calculated without all of the components being loaded, but I think there's probably a better way to do this.

JoeShep:
Does `onload`work with lazy loaded components? If not, your timer seems as good as any option.
Any other ideas, @jcreamer898 ?

jcreamer898:
:+1:

PR #155, opened 2015-10-16
Make char limit higher for sidebar items

JoeShep:
Changing the map list items titles from the POI name to the stack name makes the titles longer in many cases. That, and recent design tweaks that left room for longer titles, led to this change. No big woop.


thomasthesecond:
:+1:

PR #158, opened 2015-10-16
Hide sub nav when window is scrolled too fast on mobile

thomasthesecond:
There's an issue on mobile devices when the window is scrolled too fast to the top, the sub nav doesn't have time to unstick itself in time and appears above the header and masthead. To "fix" this bug, I just added a z-index on the masthead one higher than the content area and added a white background to the header. This way, if the sub nav doesn't unstick in time, at least it's hidden.

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -7,6 +7,10 @@
   position: relative;
   z-index: 100;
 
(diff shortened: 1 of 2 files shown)


edubkendo:
thumbsup1

PR #163, opened 2015-10-21
Run interest images through resrc.it

jcreamer898:
Should hopefully kill the massive images we have going in right now.


edubkendo:
goodjob

PR #164, opened 2015-10-21
Slightly refactor article allowing it work better in other contexts

thomasthesecond:
- Uses the `:not` pseudo-class to apply styles directly and remove override
- Puts the margin, padding, border on the top instead of the bottom
- Adds top margin to button container when it's adjacent to article list


jcreamer898:
:+1:

PR #169, opened 2015-10-26
Increase the size of the interests card

jcreamer898:

### DIFF
--- a/src/components/interests/_interests.scss
+++ b/src/components/interests/_interests.scss
@@ -1,11 +1,13 @@
 @import "../../../sass/webpack_deps";
 
 $mobile-breakpoint: 717px;
-$square-size: 190px;
+$square-size: 220px;
 $local_gutter: 30px;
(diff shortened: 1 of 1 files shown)


src/components/interests/_interests.scss
```diff
@@ -1,11 +1,13 @@
...
 
 $mobile-breakpoint: 717px;
-$square-size: 190px;
+$square-size: 220px;
 $local_gutter: 30px;
 $fluid-max: 427px;
 $fluid-min: $fluid-max + 1;
 $desktop-size: 1170px;
+// When all 9 cards are up
+$max-card-width: 1395px;
```
thomasthesecond: Should you use REMs on these values? Also, could `$local_gutter` be deleted and use `$gutter` instead since they're both the same value? If not, could we hyphenate the name to be like the others?

JoeShep: Seconded.

PR #170, opened 2015-10-27
Add click event to map markers

JoeShep:
Up to now, the user was unable to click on the map marker to load the "about" content in the map sidebar. You had to mouse up to the pop-up and click on that. This was a problem when markers were closely clustered because it was easy to accidentally hover over a different marker in the process. It also was a poor user experience to be able to hover, but not click, on the map marker. Thank you, @jcreamer898, for helping to get this working as it should


jcreamer898:
Seems legit.

PR #171, opened 2015-10-27
Things to do See More refactor

JoeShep:
Now that Top Experiences uses a slider instead of linking away from the highlight page, the decision was made to remove the link below the cards and add a visual tweak to indicate the slide show experience. 

Works across all screen widths, but could use a seasoned CSS master's eye for input on making the style code cleaner. I'm not thrilled with all the media queries.

screen shot 2015-10-27 at 10 37 50 am


JoeShep:
@thomasthesecond No rush, but would love your :eyes: when you have time.

thomasthesecond:
I didn't forget about this, I should have some time to look thru it this afternoon.

src/components/things_to_do/things_to_do.hbs
```diff
@@ -3,17 +3,17 @@
 <article class="ttd" data-lp-initial-cards="{{cards}}">
   <h2 class="ttd__heading">{{title}}</h2>
 
-  <ul class="ttd__list js-ttd-list">
-  </ul>
+  <div class="ttd__list">
```
thomasthesecond: I'd use a `js-` class here, since it's targeted in the JS. Something like `js-ttd-list-container` maybe?

src/components/things_to_do/things_to_do.js
```diff
@@ -19,6 +19,10 @@ class ThingsToDo extends Component {
       return;
     }
 
+    if (cards.length > 4) {
+      this.$el.find(".ttd__list").addClass("has-more");
```
thomasthesecond: See comment above about `js-` class on this element.

thomasthesecond:
Made some small adjustments. @JoeShep will add a z-index to the title so that it's above the shadow and then good to merge.

PR #177, opened 2015-10-29
As much as it pains my soul...clamp.

jcreamer898:
When in doubt, just clamp it...


JoeShep:
Nooooooooooooo....ok
:+1:

PR #178, opened 2015-10-30
Make the map load async

jcreamer898:
We don't need to have the entire first load of the map download initially. 
Getting it with ajax greatly increases the page load time.


edubkendo:
haruhithumbsup LGTM but dunno if maybe @JoeShep should also take a look since maps.

PR #188, opened 2015-11-09
Update framework for articles

thomasthesecond:
This update introduces [Susy](http://susy.oddbird.net/) to Rizzo Next.

Rizzo Next is lacking a grid system. Susy fills that gap by providing the tools to allow us to create grids quickly and accurately. Susy is also very flexible because it allows for multiple layouts.

For example, Susy allows for the creation of a 3-column layout very easily, like so:

```scss
// Based on a 12-column grid

.wrap {
  @include container;
}

.left-sidebar {
  @include span(2);
}

.main {
  @include span(6);
}

.right-sidebar {
  @include span(4);
}
```

Susy's documentation can be found [here](http://susydocs.oddbird.net/en/latest/).

Note that a small breaking change has been made (see 6b321bb): the `container` mixin has been renamed to `container-padded` as Susy's `container` mixin clashes. However, all of the functionality that exists within the old `container` mixin has been retained for the new `container-padded` mixin. Also, since Susy's `container` mixin does not include horizontal margins or padding, the `container-padded` mixin is able to compliment Susy's `container` mixin nicely by providing us the ability to pad edges of the container when necessary. This means that the `container-padded` mixin is usually best suited for the outermost container, whereas Susy's `container` mixin would be best used for inner containers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "sassdoc-theme-neat": "0.0.2",
     "sinon": "^1.17.1",
     "style-loader": "^0.13.0",
+    "susy": "^2.2.6",
     "webpack": "^1.12.2"
   }
 }
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -32,7 +32,7 @@ body {
 }
 
 .container {
-  @include container();
+  @include container-padded();
 }
 
 .hide-text {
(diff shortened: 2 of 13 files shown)


jcreamer898:
!()[http://i.imgur.com/XJyGY.gif]

PR #190, opened 2015-11-10
Reduce TTD title z-index so it doesn't conflict with fixed sub-nav

JoeShep:
before:
1llup

after:
screen shot 2015-11-10 at 11 26 08 am


thomasthesecond:
:+1:

PR #191, opened 2015-11-11
Add Continent navigation to masthead

JoeShep:
Creates a new feature within Masthead that allows the user to navigate to the previous or next continent, in the same order as the list of continents in the nav bar drop down. 

Relies on https://github.com/lonelyplanet/destinations-next/pull/488

screen shot 2015-11-11 at 1 54 14 pm


jcreamer898:
Looks great.

PR #192, opened 2015-11-12
Add min-height to food and drink

JoeShep:
In some edge cases, taller food and drink images were cut off when a place had only one or two food/drink POIs. Adding a min height to the component above 960px helps keep the container from collapsing to too narrow a space to fit the image.
Before:
screen shot 2015-11-12 at 2 33 37 pm
After:
screen shot 2015-11-12 at 2 30 33 pm


jcreamer898:
Me likey.

PR #193, opened 2015-11-12
Strapline bugfixes

JoeShep:
In cases where a city has more images than strap lines, the masthead text would jump to a different height when transitioning from images with strap lines to those without. This fix keeps the text in place whether `.strapline` is empty or not. I have tested the change against cities with no empty strap lines, continents, and countries. No behavior changes found in desktop or mobile views.

Also  Reduce spacing between masthead title and strapline
per https://trello.com/c/ImHr3Ell/465-title-and-strapline-tweek


jcreamer898:
:+1:

PR #194, opened 2015-11-13
Things to do updates

jcreamer898:
Hide previous button until more has been clicked.
Hide more button when at the end.
Allow for swipe on things to do component.
A few eslint updates.

### DIFF
--- a/src/components/things_to_do/_things_to_do.scss
+++ b/src/components/things_to_do/_things_to_do.scss
@@ -20,7 +20,7 @@
     margin: 0 (-$gutter/8) $gutter;
     list-style: none;
     font-size: 0;
-    text-align: center;
+    text-align: left;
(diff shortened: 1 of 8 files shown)


src/components/things_to_do/_things_to_do.scss
```diff
@@ -91,6 +101,65 @@
...
+
+      @media (min-width: $min-960) {
+        width: 2.5rem;
+      }
+    }
+
+    .has-more--left:after {
+      content: "";
+      display: block;
+      height: 120%;
```
JoeShep: Have a bit of repeated styles shared between `.has-more--left:after` and `--right:after`. Could be moved to a shared block, but not critical.

JoeShep:
Other than comment above: :shipit:

PR #200, opened 2015-11-17
Add error handling for TTD

jcreamer898:
Since we're doing this w/ ajax now, just want to make sure we cover our bases in case something 'splodes.


edubkendo:
isaac-and-miria-thumbsup

PR #204, opened 2015-11-17
Hide subnav, return to previous experiences

jcreamer898:
Return the user to previous index of experiences.
Make sure subnav element is removed if there aren't any experiences.


edubkendo:
thumbsup1

PR #205, opened 2015-11-17
Update articles "read more" link

JoeShep:
Some places don't have any articles, only related. Link changed to be less specific.


thomasthesecond:
:+1:

PR #208, opened 2015-11-18
Bump node-sass to fix includePaths

jcreamer898:
I don't *love* the *~* syntax for the `sass-loader` because it couples our sass to WebPack.

There was a bug in `node-sass` that this upgrade fixes so that the `includePaths` option works as expected.


thomasthesecond:
:+1:

PR #211, opened 2015-11-18
Build ads into article experience

thomasthesecond:
- Add jquery.dfp dependency to package.json
- Add JavaScript for ads to core; rewrite in ES6
- Add styling for ad component


jcreamer898:
This looks as good as it could for now. Well done. We can revisit later to make it more easy to understand. Good job on the conversion to ES6!

PR #221, opened 2015-11-20
Destinations beta opt-out banner

JoeShep:
Adds a banner to top of page that allows Destinations Next beta users to opt out and return to the original experience. Also has a 'close' button that will hide the banner and set a cookie to prevent showing the banner again.


edubkendo:
haruhithumbsup

PR #222, opened 2015-11-20
Override related guidebook container max-width; fix iOS Safari bug

thomasthesecond:


edubkendo:
isaac-and-miria-thumbsup

PR #223, opened 2015-11-20
Fix iOS sidescrolling bug

thomasthesecond:
- Remove CSS code that sticks image to window edge; need to revisit


edubkendo:
kyoko_thumbs_up

PR #226, opened 2015-11-22
Fix bugs in article body content

thomasthesecond:


jcreamer898:
Looks good. You can probably go ahead and turn on those events if you want and we'll just let Luc know to start grabbing them.

thomasthesecond:
Ok cool, will do. I think I saw some JS errors when the tracking code wasn't commented. I'll look into it and add the code back.

PR #227, opened 2015-11-22
Remove reveal animation on div. Add to text;

JoeShep:
This tweaks the styling for the beta opt-in/out banner. It was decided that having the banner animate open for every new page load would be a poor user experience. Animating the opacity of the text helps mitigate the "pop-in" effect from rendering on the client side a little bit.

### DIFF
--- a/src/components/alert/_alert.scss
+++ b/src/components/alert/_alert.scss
@@ -4,7 +4,7 @@
   $alert-font-color: #fff;
   max-height: 0;
   overflow-y: hidden;
-  transition: max-height 600ms;
+  transition: max-height 500ms;
 
   &--is-visible {
     max-height: 6.5rem;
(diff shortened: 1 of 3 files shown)


src/components/alert/_alert.scss
```diff
@@ -4,7 +4,7 @@
   $alert-font-color: #fff;
   max-height: 0;
   overflow-y: hidden;
-  transition: max-height 600ms;
+  transition: max-height 500ms;
```
thomasthesecond: Should we use the `$animation-speed` variable, which is 400ms? I think it's ok not to use it, but since it's only a 100ms difference, I wanted to ask.

src/components/alert/_alert.scss
```diff
@@ -31,11 +31,17 @@
...
     color: $alert-font-color;
     font-size: 1.4rem;
+    opacity: 0;
+    transition: opacity 100ms;
     padding: {
       top: 2.2rem;
       bottom: 2.2rem;
     }
 
+    &--is-visible {
```
thomasthesecond: For stateful classnames, I'd recommend not using the block name prefix: `&.is-visible`.

thomasthesecond:
Just a couple of comments. Also, should we hide the alert completely if there's no JS?

```
.alert {
  .no-js & {
    display: none;
  }
}
```

Probably not absolutely necessary, though.

Otherwise, :+1:

JoeShep:
Thanks for the :eyes: The stateful classes is a good one to know. Makes sense.

PR #228, opened 2015-11-23
Update position of social share popup menu for small screen

thomasthesecond:
This update moves the menu to the right of the button for small screens.


thomasthesecond:
Good catch, thanks!

PR #229, opened 2015-11-23
Reposition nav user item dropdown so it fits on screen

JoeShep:
This = bad:
screen_shot_2015-10-21_at_10 41 16_am

This = good:
screen shot 2015-11-22 at 8 30 35 pm

This PR = make it good


jcreamer898:
:shipit:

PR #232, opened 2015-11-23
Uncomment track code and rename events

thomasthesecond:


edubkendo:
isaac-and-miria-thumbsup

PR #236, opened 2015-11-23
Fix iPad scrolling issue in maps panel

JoeShep:
Having the footer in the panel div caused issues in iPads where the footer would not stay fixed to the bottom. Having the footer rendered by handlebars instead of hardcoding it into the template allowed the footer to be 'outside' of the panel div and overcomes the scrolling problem.


jcreamer898:
:+1:

PR #237, opened 2015-11-23
fix top places layout bug

JoeShep:
The top places list of cities was adding a top margin at the wrong breakpoint.


jcreamer898:
:+1:

PR #238, opened 2015-11-23
Update fade-edge mixin

thomasthesecond:
`transparent` can't be used as a color value in a `linear-gradient` in iOS Safari, so it has been replaced with a `rgba`.

Also added a `color` param to the mixin.


JoeShep:
Glad it wasn't a huge pain in the arse fix. :+1:

PR #240, opened 2015-11-24
Tweak hotels btn css to work in IE

JoeShep:
The margin on the search button was exactly one pixel too wide to work in IE. Reducing it keeps it from jumping below the form inputs. Also needed to add an IE-specific style to hide the HTML5 select dropdown arrow icon.

Before:
10qpb
After:
screen shot 2015-11-24 at 11 56 36 am


thomasthesecond:
:+1:

PR #242, opened 2015-11-24
Remove lazy loading of alert text animation to fix articles bug

JoeShep:
Alert text wouldn't show on articles. Lazy load logic was biased to the DN highlight page. Quick fix is to remove lazy load of opacity animation


jcreamer898:
:+1:

PR #245, opened 2015-11-30
Add matchMedia utility to core

thomasthesecond:
This utility uses the [window.matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) method to match a media query string. It will allow conditional loading of components or pieces of components based on a media query.

Browser support is pretty good, but it will need to be polyfilled for IE 9 and Opera Mini 8. See http://caniuse.com/#feat=matchmedia

Usage is documented in the code, but it's very straight forward:

```js
import matchMedia from "../../core/utils/matchMedia";

matchMedia("(min-width: 720px)", (query) => {
  if (query.matches) {
    // do this
  } else {
    // do that
  }
});
```

### DIFF
--- a/src/core/utils/matchMedia.js
+++ b/src/core/utils/matchMedia.js
@@ -1,0 +1,26 @@
+/**
+ * Checks against a specified media query string. Adds a listener for
+ * `matchMedia` and runs a callback function when the media query matches.
+ * @param  {String}   query    Media query to listen for
+ * @param  {Function} callback Callback function to run when the query
+ *                             matches.
+ * @example
+ * import matchMedia from "../../core/utils/matchMedia";
+ *
+ * matchMedia("(min-width: 720px)", (query) => {
+ *   if (query.matches) {
+ *     // do this
+ *   } else {
+ *     // do that
+ *   }
+ * });
+ */
+export default function(query, callback) {
+  let query = window.matchMedia(query);
+
+  if (typeof callback === "function") {
+    query.addListener(callback);
+
+    callback(query);
+  }
+}
--- a/src/core/utils/matchMedia.js
+++ b/src/core/utils/matchMedia.js
@@ -16,11 +16,11 @@
  * });
  */
 export default function(query, callback) {
-  let query = window.matchMedia(query);
+  let media = window.matchMedia(query);
 
   if (typeof callback === "function") {
-    query.addListener(callback);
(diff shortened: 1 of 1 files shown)


thomasthesecond:
What would be the best way to polyfill this? I prefer https://github.com/weblinc/media-match but it doesn't look like it supports NPM. However, this one from Paul Irish does https://github.com/paulirish/matchMedia.js/.

Should the polyfill be imported into the utility itself, or should it somehow be loaded via Webpack if needed?

thomasthesecond:
Paul Irish version seems to be more up-to-date.

jcreamer898:
:+1:

thomasthesecond:
I looked into the `Modernizr.mq` method, but it doesn't really do what I'd like it to do. `Modernizr.mq` returns a boolean on load and there is no listener, which means I'd have to use a `window.resize` to always check if the breakpoint matches. Even with a debounce, the `window.resize` calls my method many times. I suppose I could stop calling it as soon as I get a `true` or `false`. Seems like extra unnecessary work that `window.matchMedia` does already.

Using the native `window.matchMedia` is a little nicer because it provides a built-in listener and I'd prefer to build for the future when possible.

I think `Modernizr.mq` works well enough where it's being used, but for what I need, it's not really working. Is there any reason we can't keep both methods around for now? I suppose there could be some confusion as to which method to use, but I'd always prefer and recommend `window.matchMedia`.

thomasthesecond:
http://stackoverflow.com/a/21558678

jcreamer898:
Alrighty then, go ahead w/ what you were doing and bring in Irish's polyfill. I wanna make sure we test this in at LEAST IE11 and Edge, FireFox, Chrome, and Safari. Possibly an Android and iPhone too. Just to make sure we don't bork the page.

thomasthesecond:
`window.matchMedia` is an API for finding out whether or not a media query applies to the document. It returns a new [MediaQueryList](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) object representing the parsed results of a specified media query string. Read more about `window.matchMedia` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia).

The required polyfill can be found on the related Destinations Next PR here: https://github.com/lonelyplanet/destinations-next/pull/553

PR #246, opened 2015-11-30
Article POI callout

thomasthesecond:
The POI callout component is currently only used in articles. Articles that mention (and link to) a POI on-site will use this component to show more information about the POI in a tooltip-like menu when the link is hovered.

The article body component has been updated as well in order to accommodate the POI callout. POI data is loaded via AJAX in the article body component, cached and then passed to the POI callout component. A promise is used to make sure the data is ready. Lastly, matchMedia is used to conditional initialize or destroy the POI callout inside of the article body.

### DIFF
--- a/src/components/poi_callout/poi_callout.js
+++ b/src/components/poi_callout/poi_callout.js
@@ -1,0 +1,227 @@
+import { Component } from "../../core/bane";
+import $ from "jquery";
+import $clamp from "clamp-js/clamp.js";
+import debounce from "lodash/function/debounce";
+
+export default class PoiCalloutComponent extends Component {
+  /**
+   * Render the POI callout
+   * @return {jQuery} Returns the poi-callout element
+   */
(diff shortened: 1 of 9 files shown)


src/components/poi_callout/poi_callout.js
```diff
@@ -0,0 +1,227 @@
...
+
+    if (poiData.image) {
+      // TODO: Do this on the Ruby side
+      image = poiData.image.replace("http://media.lonelyplanet.com/", "");
+      src = `https://lonelyplanetimages.imgix.net/${image}?w=210&h=95&fit=crop`;
+    } else {
+      src = "";
+    }
+
+    this.$callout
```
jcreamer898: I think what make your life easier here is just re-rendering the template again?

thomasthesecond:
Verbal OK from @jcreamer898.

PR #247, opened 2015-12-01
Updating how we send events to tealium

jcreamer898:
Add an option to not return the value as data.
Send data to Tealium's procEvent.
A quick guidebook change is sneaking in here too.


edubkendo:
goodjob

benfalk:
Does it make sense to squash these two commits together into one?  This looks like the "Old One Two"

PR #249, opened 2015-12-02
Article component revisions

thomasthesecond:


sass/_mixins.scss
```diff
@@ -461,11 +461,40 @@
...
+  display: block; // Fallback for non-webkit browsers
+  display: -webkit-box;
+  font-size: $font-size;
+  height: ($font-size * $line-height * $lines-to-show); // Fallback for non-webkit browsers
+  line-height: $line-height;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: $lines-to-show;
+}
```
JoeShep: Is this a replacement for the clamp js package? If not, seems a little confusing to have two clamping processes in the same code base. Any thoughts, @jcreamer898?

JoeShep: Otherwise, :+1:

jcreamer898: Agreed.

thomasthesecond:
I can see where there'd be confusion. The reason I tried a CSS method was because that POI data is being loaded dynamically and the JS method to clamp seemed to work half the time. Not sure if it was because of the dynamically added content or not. Also, I thought a CSS-only solution would be better for progressive enhancement. But anyway, I think we just decided to truncate on the back-end, so I'll remove the mixin. Thanks for the feedback!

PR #252, opened 2015-12-04
Remove Tabs component from articles; Add standard heading

JoeShep:
With the removal of news articles for the time being, it was decided to remove the tabs altogether and have the articles component display the same title design as other components.

Before:
screen shot 2015-12-04 at 12 19 50 pm

After:
screen shot 2015-12-04 at 12 19 39 pm


jcreamer898:
Works for me.

PR #253, opened 2015-12-04
Tweak 'see more' icon height; Change "tours" to "adventures

JoeShep:
Changing the 'branding' of the tours component to "adventures" is an attempt to boost user interaction. Also threw in a tweak to the right arrow on "see more" links to better center them vertically


jcreamer898:
:+1:

PR #257, opened 2015-12-09
"Search _____ and beyond" comes from `search.title`

mkunkel:
Now that DN serves more than just place pages, the title used in the search bar should be derived from a variable that is agnostic to the type of page being rendered.


jcreamer898:
:+1:

PR #259, opened 2015-12-11
Adding masthead video

jcreamer898:
This commit enables some videos in the masthead.
Adds a jsconfig.json file for vscode.
Add a cover video util.


src/components/masthead/masthead.hbs
```diff
@@ -3,8 +3,18 @@ By use of this code snippet, I agree to the Brightcove Publisher T and C
 found at https://accounts.brightcove.com/en/terms-and-conditions/.
 -->
 <article class="masthead masthead--{{type}}{{#if images}} masthead--has-images{{else}} masthead--no-images{{/if}}">
+  {{#if video}}
+  <div class="masthead__video-container js-video-container">
+    <div class="masthead__video-wrap">
+      <video class="masthead__video js-video" autoplay loop preload="auto">
+        <source src="{{video.source}}" type="video/mp4">
```
thomasthesecond: Should the `type` be dynamic too?

src/components/header/_header.scss
```diff
@@ -18,6 +18,8 @@
     top: 0;
     left: 0;
     right: 0;
+    
+    background: linear-gradient(to bottom, rgba(0,0,0,0.2), transparent);
```
thomasthesecond: I don't think iOS Safari works with `transparent`. I tried something similar and had to fix it [here](https://github.com/lonelyplanet/rizzo-next/pull/238). So I'd change `transparent` to `rgba(0,0,0,0)`.

thomasthesecond:
Just two small comments but otherwise :+1:

PR #262, opened 2015-12-12
Remove ads from article experience

thomasthesecond:
- Ad assets are not secure and will not load
- Also pre-load next article later


jcreamer898:
Bummer.

PR #265, opened 2015-12-14
Load article share component in article component

thomasthesecond:
And also load share component for new articles loaded with AJAX

Create action_sheet.js and update index.js


src/components/action_sheet/action_sheet.js
```diff
@@ -0,0 +1,85 @@
...
+    let $el = $(event.currentTarget);
+
+    let width = 550,
+        height = 420,
+        winHeight = $(window).height(),
+        winWidth = $(window).width(),
+        left,
+        top;
+
+    let title = $el.closest(".article").find("meta[itemprop=\"headline\"]")[0].content,
```
jcreamer898: You *might* want to check for the existence of the tag before assuming it has content.
```js
let $title = $el.closest(".article").find("meta[itemprop=\"headline\"]");

let title;
if ($title.length) {
  title = $title[0].content;
}
```

src/components/action_sheet/action_sheet.js
```diff
@@ -0,0 +1,85 @@
...
+        url = window.location.href,
+        network = $el.data("network");
+
+    let tweet = `${urlencode(title)} ${urlencode(url)} via @lonelyplanet`;
+
+    left = Math.round((winWidth / 2) - (width / 2));
+    top = 0;
```
jcreamer898:
```js
top = winHeight > height ? Math.round((winHeight / 2) - (height / 2)) : 0;
```

jcreamer898:
Few comments, otherwise :+1:

PR #266, opened 2015-12-15
Update meta data for newly AJAX-loaded articles

thomasthesecond:
This pull request updates meta data for articles as they're loaded with AJAX and scrolled to. This also fixes an error on the `join` method where sometimes `interests` is a string instead of an array.


jcreamer898:
:+1:

PR #268, opened 2015-12-15
Put ads back into article experience

thomasthesecond:
- Remove initialization of lp.ads, moved to DN
- Remove updateAdLayers and updateAdPath methods
- Add 940 x 40 to ad sizes array
- Update method slugifyPlaceDataForAds (renamed, used for lp.ads not lp.article)


jcreamer898:
:+1:

PR #269, opened 2015-12-15
Add event for new AJAX page view

thomasthesecond:
Send `title` to `trackEvent` method and track an event when a virtual pageview is triggered.


jcreamer898:
:+1:

PR #270, opened 2015-12-16
Update events and how they're tracked for articles

thomasthesecond:
Refactored analytics.js a bit so that a meaningful string can be passed instead of an object, as per Google's docs.

eventCategory - Typically the object that was interacted with (e.g. 'Video')
eventAction - The type of interaction (e.g. 'play')
eventLabel - Useful for categorizing events (e.g. 'Fall Campaign')

### DIFF
--- a/src/components/article/article.js
+++ b/src/components/article/article.js
@@ -39,6 +39,10 @@ export default class ArticleComponent extends Component {
     this._setFirstArticle();
     this._createInitialListOfArticles();
     this._loadFirstAd();
+
+    this.events = {
+      "click .article-pagination__item": "_trackArticlePagination"
+    };
   }
 
   _slugifyPlaceDataForAds() {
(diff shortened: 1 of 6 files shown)


src/components/article/article.js
```diff
@@ -538,15 +549,21 @@ export default class ArticleComponent extends Component {
...
       .build();
   }
 
-  @track("ad page load impression");
+  @track("article ad impression load");
   _loadFirstAd() {
     let adSlots = [];
     let $adUnit = $("#adunit-0");
 
+    if (this.$window.width() >= 728) {
```
jcreamer898:
```js
$adUnit.data("sizeMapping", this.$window.width() >= 728 ?  "leaderboard" : "mpu") ? 
```

jcreamer898:
one comment, then :+1: 

Just double and triple check the existing events all still work (top places, hotels search, etc).

PR #271, opened 2015-12-18
Add a quick fix to wait for utag before tracking first ad impression

thomasthesecond:
The problem is that `utag` is undefined when the page loads because it is loaded at the bottom of the page. Napkyn says "ideally [utag] `init` should be as close to the opening `<body>` tag as possible."

Therefore, I've added a `setTimeout` to wait until `utag` has been defined and then load the first add and track the impression. There may be a better way to do this, but for now, I believe this will do the trick.


JoeShep:
Clever, clever. :+1:

PR #272, opened 2015-12-21
Improve scroll to next feature for article experience

thomasthesecond:
This update changes how the scroll to next feature works.

The calculations for loading articles has been improved. Instead of using each article's height and top offset, calculations using the document and window height are used. This allows the new article to be loaded near the bottom of the page and it removes the need to use `setTimeout` to get calculations.

Some updates have been made in order to animate the new article into place. Mostly just adding classnames and using the `waitForTransition` method.

A promise has been created for the next article AJAX call.

Styling updates have been made to the article and article loader in order to transition in/out when loaded/removed.

Lastly, made a small styling update to the ad component so that collapsed ads aren't shown.

### DIFF
--- a/src/components/ads/ads.scss
+++ b/src/components/ads/ads.scss
@@ -39,7 +39,7 @@
 
 /// Adunit
 
-.adunit--leaderboard > div {
+.adunit {
   padding-bottom: 1.8rem;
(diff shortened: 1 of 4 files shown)


jcreamer898:
:+1:

PR #273, opened 2015-12-21
Update styles for article

thomasthesecond:
This fixes various bugs and styling issues present within the new article experience.

### DIFF
--- a/src/components/article_body/article_body.scss
+++ b/src/components/article_body/article_body.scss
@@ -106,9 +106,9 @@
       }
     }
 
-    li a:not([data-poi-slug]),
-    p:not(.stack__article__image-container) a:not([data-poi-slug]),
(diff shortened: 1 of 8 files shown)


edubkendo:
kyoko_thumbs_up

src/components/article_body/article_body.scss
```diff
@@ -82,9 +96,20 @@
...
+      font-size: 1.4rem;
+      line-height: (24 / 14);
+
+      @media (min-width: $min-720) {
+        font-size: 1.6rem;
+        line-height: (24 / 16);
+      }
+    }
+
+    li a:not([data-poi-slug]),
```
mkunkel: These should probably change to `data-callout-slug`

thomasthesecond: Good catch, thanks. I'll change it here and in the JS.

PR #275, opened 2015-12-29
Change :before and :after positioning to better hide TTD sliding imgs

JoeShep:
On certain wide displays the top experiences slideshow was not properly hiding the images as they slid in and out of view. This fix simplifies the CSS and does a better job of masking the moving images.


thomasthesecond:
:+1:

PR #279, opened 2016-01-04
Update fonts for webfont loader

thomasthesecond:
This pull request makes some changes to fonts and font variables.

#### Rename custom fonts

These names are now representative of the each font's actual name. Ideally, the fonts should be declared by their actual names and not a keyword. "Main" and "secondary" are too ambiguous. What would happen if a third font was added? Consider the approach taken by Typekit and Google Fonts.

Anyone looking at the code now will be able to identify what font is being used where without having to know that a specific keyword matches a certain font.

This update will also benefit the webfont loader because the fonts need to be declared within the loader, so it'll help if the names are descriptive.

#### Update font variables

A number of updates have been made to how font variables are defined and some variables have been deprecated.

1. `$sans-serif` and `$serif` variables mapped to `$font-family` and `$font-family-special`, respectively. There isn't a need to redefine these variables with a different name. Therefore, `$sans-serif` and `$serif` have been updated to declare a font stack. These stacks can be used independently or appended to custom fonts. They use the generic variable names `$sans-serif` and `$serif` because they are to be treated as generic fonts that aren't technically part of the design but to be used as fallbacks.
2. `$font-family` and `$font-family-special` have been deprecated. These variable names are rather ambiguous and don't describe which font is being used, which leads to #3.
3. `$benton-sans` and `$miller-daily` variables have been created. Since each custom font's name was changed to be more descriptive, these new variables serve the same purpose. With these new variables, it will be clear what font is being requested.
4. `$light`, `$medium`, `$semibold` have been replaced and deprecated. Again, these variables only serve to redefine variables that have already been created. `$font-weight-light`, `$font-weight-medium`, `$font-weight-semibold` have been kept because they are more descriptive. Deprecated variables should be removed in a future release.

#### Update files to add new variables and replace deprecated ones

Of course this needs to be done so the stylesheets don't break.

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -8,7 +8,7 @@ body {
   position: relative;
 
   line-height: $line-height;
-  font-family: $sans-serif;
+  font-family: $benton-sans;
   font-size: $base-size;
   color: $font-color;
 
--- a/sass/_fonts.scss
+++ b/sass/_fonts.scss
@@ -1,14 +1,14 @@
 /*
-This CSS resource incorporates links to font software which is 
-the valuable copyrighted property of WebType LLC, The Font Bureau, 
-and/or their suppliers. You may not 
-attempt to copy, install, redistribute, convert, modify or reverse 
-engineer this font software. Please contact WebType with any 
-questions: http://www.webtype.com 
+This CSS resource incorporates links to font software which is
+the valuable copyrighted property of WebType LLC, The Font Bureau,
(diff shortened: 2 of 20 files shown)


thomasthesecond:
Related: https://github.com/lonelyplanet/destinations-next/pull/596

jcreamer898:
Fix dem' conflicts then :+1:

PR #285, opened 2016-01-06
Support passing in a path to a JSON config

jcreamer898:
Will work on the spec for this later, but for now, I've added support for passing in a path to a configuration file. You could then pick and choose which components to render.

This also changes the default header class to `header-normal`


thomasthesecond:
:+1:

PR #287, opened 2016-01-06
Update sights component to match design

thomasthesecond:
This pull request updates the sights component to match the design, but also has been refactored to remove complexity. The layout now uses flexbox.

One thing that I don't like is how `sights__item` is repeated 3 times in sights.hbs. I've created sights_item.hbs, but I can't include the partial inside of an `{{#each}}`. 

What I would like to do is this:

```handlebars
{{#each first_column}}
  {{> "components/sights/sights_item" }}
{{/each}}
```

But that causes the following error in Ruby:

```
undefined method `render' for #<OpenStruct …
```

Any suggestions?

### DIFF
--- a/sass/elements/_buttons_mixins.scss
+++ b/sass/elements/_buttons_mixins.scss
@@ -30,8 +30,8 @@
 
   [class*="icon-"] {
     display: inline-block;
-    font-size: .6rem;
-    margin-left: .6rem;
+    font-size: 6px;
+    margin-left: 6px;
     transform: translate3d(0, 0, 0);
     transition: transform 500ms;
     vertical-align: 20%;
--- a/sass/typography/_links_mixins.scss
+++ b/sass/typography/_links_mixins.scss
@@ -32,9 +32,9 @@
   [class*="icon-"] {
     backface-visibility: hidden;
     display: inline-block;
-    font-size: .6rem;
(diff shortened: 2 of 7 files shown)


thomasthesecond:
I'll load the sights_item template client-side.

thomasthesecond:
Related: https://github.com/lonelyplanet/destinations-next/pull/601

sass/elements/_buttons_mixins.scss
```diff
@@ -30,8 +30,8 @@
 
   [class*="icon-"] {
     display: inline-block;
-    font-size: .6rem;
-    margin-left: .6rem;
+    font-size: 6px;
```
jcreamer898: What's up w/ no moar rem's here?

jcreamer898:
Looks good. Just make sure we get some good testing done on no flexbox browsers.

sass/elements/_buttons_mixins.scss
thomasthesecond: Mentioned in commit https://github.com/lonelyplanet/rizzo-next/commit/e10a73b6a8ee0939b0b660bda282b1c566edcc86.

REM: http://cl.ly/2M220N1m243s
Pixels: http://cl.ly/0G3Y1Z1B0E0M

PR #290, opened 2016-01-08
Improve things to do component

thomasthesecond:
There were some layout and design issues with the component. There was also some side scrolling on iOS. Additionally, there was some unnecessary complexity in the CSS code which has been removed. Lastly, the markup for the more and less buttons have been removed from the JS and placed within a handlebars template.


jcreamer898:
Me likey.

jcreamer898:
Tested in a few browsers?

thomasthesecond:
I need to do that and also add a fallback for flexbox.

PR #292, opened 2016-01-11
Add in babel polyfill to build, update link for messages

jcreamer898:
Hopefully global nav will work in IE9+ with this update.


thomasthesecond:
:+1:

PR #293, opened 2016-01-11
Fix missing fallback image for sights

thomasthesecond:
Import topic partial and reset `position` of `topic__image` (`sights__image`) to `static`.


jcreamer898:
nice

PR #295, opened 2016-01-11
Move conditional from inside of sights component to outside

thomasthesecond:
This is done so that if there are no sights, then the entire component is not rendered.


jcreamer898:
Nice.

PR #299, opened 2016-01-13
Namespace search component with "lp-" prefix

thomasthesecond:
This is being done to avoid naming conflicts while Rizzo Next is used alongside of Rizzo.

### DIFF
--- a/src/components/overlay/index.js
+++ b/src/components/overlay/index.js
@@ -6,10 +6,8 @@ import "./index.scss";
 
 class Overlay extends Component {
 
-  initialize(options = {}){
-    this.options = Object.assign(options, {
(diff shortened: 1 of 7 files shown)


jcreamer898:
My god that was fast.

src/components/overlay/index.js
```diff
@@ -8,7 +8,7 @@ class Overlay extends Component {
 
   initialize(options = {}){
```
jcreamer898: You SHOULD be able to use default params here...

```js
 initialize(options = {
  preventScroll: true
}){
```

WunderBart:
:boom: :+1: \o/

PR #303, opened 2016-01-14
Update header search focus styles

thomasthesecond:
Removed focus outline, but added hover styles to active and focus states to retain some trace of accessibility. May need to be improved upon.


jcreamer898:
:+1:

PR #305, opened 2016-01-14
Fix issues with TTD aka experiences components

thomasthesecond:
Update selector from `#ttd` to `#experiences`

- This will fix the issue of the experiences section not being removed if empty.

Fix experiences notifier

- Update ttd to experiences in `subscribe` method to fix sub nav link not being removed
- Inside of `addSights` method, first check if `.sights` exists, if not, remove the closest `.segment` aka its parent


jcreamer898:
:+1:

PR #306, opened 2016-01-14
Add www. to "more" links

thomasthesecond:
Fixes an SSL error


thomasthesecond:
@jcreamer898 gave me a :+1:

PR #307, opened 2016-01-14
Remove preventDefault from handleClick method

thomasthesecond:
Allows mobile sub navigation links to click through.


jcreamer898:
:+1:

PR #310, opened 2016-01-18
Fix some issues that popped up from trackjs

jcreamer898:
I *think* utag was undefined but, "use strict" was messing up my check here.
Also, we were getting double errors because of the `console.error`


thomasthesecond:
:+1:

PR #324, opened 2016-01-21
Remove static container from narrow header

### LINKED ISSUE #323: Article header too wide
At certain window widths, the article header is too wide. This is because it has an absolute `width` but instead needs a `max-width`.

screen shot 2016-01-21 at 2 07 38 pm

thomasthesecond:
Replace with `max-width` to fix issue where header can be too wide.

Fixes #323


jcreamer898:
:+1:

PR #327, opened 2016-01-21
Replace `percent` function with Sass's native `percentage`

thomasthesecond:


jcreamer898:
:+1: after fix conflicts

PR #329, opened 2016-01-23
Clean up z-index

thomasthesecond:
- Deprecate `z-layers` mixin
- Create `$z-index` map to hold values
- Create `z` function to return value
- Update components to use new function and values


jcreamer898:
Holy crap that's awesome.

PR #331, opened 2016-01-25
Add new partial for animations

thomasthesecond:
This PR pulls animations out of component files and puts them in one file.


jcreamer898:
Awesome work.

PR #333, opened 2016-01-25
Fix POI callout image

### LINKED ISSUE #332: Broken images on POI callouts within articles
Images are broken on POI callouts that have images.

thomasthesecond:
Now that imgix is being used to set the image, the conditional and `replace` method is no longer needed.

Fixes #332


jcreamer898:
:+1:

PR #337, opened 2016-01-25
Remove rule for article images with `aligncenter` class

### LINKED ISSUE #336: Article images can become off-centered on smaller screens
When the viewport is below 840px, article images are off-centered. See https://www.lonelyplanet.com/denmark/travel-tips-and-articles/76803 for an example.

thomasthesecond:
This rule causes images to become off-center at viewports less than 840px. It can be removed because it's not doing anything to help.

Fixes #336


jcreamer898:
:+1:

PR #338, opened 2016-01-25
Use absolute URL for sights link

thomasthesecond:
The "see all sights" link needs to go to `http` instead of `https` which means the link needs to be absolute rather than relative since the new site is forcing `https`.

Fixes lonelyplanet/destinations-next#639


jcreamer898:
:+1:

jcreamer898:
:+1:

PR #340, opened 2016-01-26
Remove code that checks tt feed. Closes #316

### LINKED ISSUE #316: Remove User Feed
There's no reason to poll for messages and activity anymore as we've removed that from the global navigation. Need to remove that code.

jcreamer898:
We no longer have any thorntree stuff in the nav, so  removing this code to stop polling.


thomasthesecond:
:+1:

PR #345, opened 2016-01-26
Prevent masthead text from extending off screen

### LINKED ISSUE #334: Masthead text can extend off-screen
In the masthead component, sometimes a title is too long and can extend off-screen on smaller screens. See https://www.lonelyplanet.com/nepal/solukhumbu for example.

thomasthesecond:
Fixes #334

### DIFF
--- a/src/core/utils/fitText.js
+++ b/src/core/utils/fitText.js
@@ -1,5 +1,14 @@
 import debounce from "lodash/function/debounce";
 
+let setFontSize = ($el, textWidth, settings) => {
+  let compressor = textWidth / $el.width();
+  $el.css("font-size", Math.floor(settings.minFontSize / compressor));
+};
+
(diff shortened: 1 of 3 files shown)


src/core/utils/fitText.js
```diff
@@ -0,0 +1,56 @@
...
+
+  $.extend(settings, options);
+
+  if (!$el.find("span").length) {
+    $el.wrapInner("<span />");
+  }
+
+  let textWidth = $el.find("span").width();
+
+  function setFontSize() {
```
jcreamer898: You might want to declare these functions outside of `fitText`, I'm not entirely sure here, but I THINK the way you have this right now, you'll essentially be redeclaring these functions every time you call fitText.

jcreamer898:
One comment then :+1:

PR #346, opened 2016-01-26
Refactor map pin

### LINKED ISSUE #343: Black triangle on map callout in Chrome
pasted_image_at_2016_01_26_10_47_am

thomasthesecond:
The current implementation had `icon` classnames on the pin element itself, which caused some issues:

1. It added the icon font as the `font-family` for the pin which means that pin text was not using Benton
2. It added a background image to the pin (this was being overridden in CSS)

The solution was to create an icon element and apply the icon styles to that element.

Fixes #343

### DIFF
--- a/src/components/map/sidebar/item.scss
+++ b/src/components/map/sidebar/item.scss
@@ -1,6 +1,7 @@
 .place {
   background-color: #fff;
   cursor: pointer;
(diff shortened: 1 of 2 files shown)


jcreamer898:
:+1:

PR #347, opened 2016-01-26
Remove wait for transition to focus search input

### LINKED ISSUE #335: Auto-focus search input on mobile
On mobile, when the search button is tapped, the input should have focus when it appears.

thomasthesecond:
I have not been able to figure out why, but on iOS, the search input will not focus when it's inside of the `waitForTransition` method. Removing the method seems to have no negative effects on desktop or mobile, but the input now gets focused on mobile.

In my debugging, the `waitForTransition` method is working properly because I was able to `console.log` inside of it. I've tried `blur()`ing the button before adding `focus()` to the input, but that didn't work either.

Fixes #335

### DIFF
--- a/src/components/search/search_component.js
+++ b/src/components/search/search_component.js
@@ -93,10 +93,7 @@ class SearchComponent extends Component {
       this.$el.addClass(SearchComponent.className + "--visible");
     }, 10);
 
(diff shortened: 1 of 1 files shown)


jcreamer898:
Seems legit.

PR #349, opened 2016-01-26
Fix browser history bug with map

thomasthesecond:
Fixes lonelyplanet/destinations-next#455

Added a state object when using `pushState` and then used the `window.onpopstate` method to check the state.

Removed the `createHistory` module import. I'm not sure why that was being used instead of the native `window.history`. The `createHistory` module seemed to mess up the value of the state when it was updated with `pushState`. It would change it from a boolean to a hashed string (something like `{isOnMap: "xY9ds"}`) which would break functionality.

### DIFF
--- a/spec/unit/components/map.spec.js
+++ b/spec/unit/components/map.spec.js
@@ -12,7 +12,7 @@ let MapComponent = Injector({
   "history": {
     createHistory: function() {
       return {
-        pushState: pushState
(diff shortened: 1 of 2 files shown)


thomasthesecond:
I need to look into the `history` module a bit more and get the test to pass.

jcreamer898:
:+1:

PR #353, opened 2016-01-27
Fix superzone sizing issues. Closes #352

### LINKED ISSUE #352: Get superzone rendering correctly ads bug in progress
The thing wasn't sizing right. Added a callback for when the superzone loads.

jcreamer898:
Needed to make sure the superzone ad could take up all the space it needs.

You can now register callbacks for certain sizes.

### DIFF
--- a/src/core/ads/ad_manager.js
+++ b/src/core/ads/ad_manager.js
@@ -25,9 +25,18 @@ export default class AdManager {
 
     this.config = $.extend({}, this.defaultConfig, config);
 
+    this.adCallbacks = {
+      "1x1": "_superzone"
+    };
+
     this.subscribe();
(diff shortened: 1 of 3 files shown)


src/core/ads/ad_manager.js
```diff
@@ -25,11 +25,20 @@ export default class AdManager {
...
+    };
+
     this.subscribe();
 
     return this;
   }
 
+  _superzone($unit) {
+    $unit.removeClass("adunit--leaderboard")
+      .addClass("adunit--superzone");;
```
thomasthesecond: Double `:`

thomasthesecond:
One thing but :+1:

PR #354, opened 2016-01-27
Make sure to always request .json

jcreamer898:
For whatever reason, sometimes people are getting 404's hitting that ajax endpoint. O_o


thomasthesecond:
:+1:

PR #365, opened 2016-01-29
Add matchMedia polyfill

thomasthesecond:
For IE 9


thomasthesecond:
@jcreamer898 says :+1:

PR #368, opened 2016-02-01
Using only TrackJS for now

### LINKED ISSUE #367: Hitting airbrake rate limit
Because of some crazy JS errors we're killing our rate limit.

jcreamer898:
Closes #367. Going to Try only TrackJS for a bit.


thomasthesecond:
:+1:

Issue #372, opened 2016-02-02
Tighten up article footer, move ad


thomasthesecond:
Remove paginated links, tags. Tighten up spacing.

Load ad at the top of new article, instead bottom of current article.

PR #377, opened 2016-02-03
Update article footer

thomasthesecond:
This PR removes some things from the article footer (categories, pagination/related) and changes how the leaderboard ads load in for each new article (prepended to top of new article rather than appended to the bottom of the current). This means that the first article on load will not have a leaderboard ad.


thomasthesecond:
Resolves #372 and lonelyplanet/destinations-next#667.

jcreamer898:
:+1:

PR #378, opened 2016-02-03
State targeting

jcreamer898:
Closes lonelyplanet/destinations-next#666. Getting state targeting working.


thomasthesecond:
Would it make sense to put the `slugify` method in a separate file, maybe inside of core/utils? Otherwise :+1:

PR #381, opened 2016-02-04
Trying to hone in on trackJs errors

jcreamer898:
We're getting `[object Object]` on trackJs. Going to stringify myself and see if that helps.


thomasthesecond:
:+1:

Issue #370, opened 2016-02-01
Masthead title off centered in IE


thomasthesecond:
Versions 9, 10 and 11 (pictured).

![screen shot 2016-02-01 at 4 21 17 pm](https://cloud.githubusercontent.com/assets/1479563/12733351/dd6bf138-c8ff-11e5-97a0-ab5f7ca90f44.png)

PR #384, opened 2016-02-04
Fix for issue #370

webdeveric:
Setting box-sizing causes the text-wrap to center correctly in IE10.


jcreamer898:
@thomasthesecond there ya go!

jcreamer898:
Good catch. Thanks @webdeveric

thomasthesecond:
:+1: Thanks!

PR #388, opened 2016-02-04
Add logo component

thomasthesecond:
This PR makes the logo its own component and makes it a little more DRY.


thomasthesecond:
I've been thinking about the best way to handle implementation detail, like so:

https://github.com/lonelyplanet/rizzo-next/blob/tc-logo-component/src/components/logo/logo.scss#L23

Article here: http://csswizardry.com/2015/06/contextual-styling-ui-components-nesting-and-implementation-detail

Should we add a class like I have? Or do something like this:

```scss
.logo {
  // styles
  .header & {
    // implementation detail
  }
}
```

I'm not sure one way is better, but I'm trying to figure out what makes the most sense for us.

jcreamer898:
I'd think the parent component would control it, like you have it up there?

thomasthesecond:
I changed the implementation from

```scss
.lp-logo {
  &.in-global-header { ... }
}
```

to

```scss
.lp-logo {
  .lp-global-header & { ... }
}
```

I think it's ready to merge.

PR #389, opened 2016-02-05
Update clearfix

thomasthesecond:
- `::before` and `zoom` can be removed because they're for IE 6 and 7
- `block` instead of `table` for `display` so that margins collapse
- space in `content` for Opera


jcreamer898:
:+1:

PR #391, opened 2016-02-05
Masthead updates

### LINKED ISSUE #390: Masthead text shrinks when it shouldn't
With the new `fitText` method added in #345, the masthead text sometimes resizes when it shouldn't. Specifically, this happens when the window width starts wide and then is narrowed. The initial width calculation on the text can be wide and the text starts to get resized too soon. This only seems to happen on titles with more than one word, however, I haven't found the edge case, yet.

thomasthesecond:
Simplifies the masthead CSS a bit, adds a conditional around the breadcrumbs element to hide if there's no content and removes the sponsor ad if there's no content.

Also fixes #390.


jcreamer898:
:+1:

PR #394, opened 2016-02-05
Add legacy icons for Best in US page

thomasthesecond:
Some icons that were removed/refactored are being used on the Best in US page. They've been re-added for now.


jcreamer898:
:+1:

PR #397, opened 2016-02-09
Setup cookie compliance

jcreamer898:
Closes lonelyplanet/destinations-next#694.
Make alert more re-usable.
Turn on alert for cookie compliance in EU.
Change map-alert class for better specificity.


thomasthesecond:
:+1:

PR #400, opened 2016-02-10
Change pushState to replaceState

thomasthesecond:
Also, refactor `_scrollToNextArticle` method and remove `$footer`.

Fixes lonelyplanet/destinations-next#669


jcreamer898:
:+1:

PR #402, opened 2016-02-10
Remove the resrc.it url from sights component

mkunkel:
A change in DN will include full URLs for these images, resulting in broken images. Also, we should be switching over to imgix anyway.


jcreamer898:
:+1:

PR #407, opened 2016-02-10
youtube ios fix

oczywisty:
There was youtube autoplay problem on iOS devices. Now it autoplays only on the devices that support that (detection on youtube's part). I removed the javascript youtube player dependency, now it's pure html approach, should work everywhere ;)

Now it's also possible to click on the youtube's movie name to go directly to youtube.

I also removed some functions duplicates, which I missed when I was rebasing 2 weeks ago.


thomasthesecond:
:+1:

PR #408, opened 2016-02-11
set negative z-index for mobile-navigation before initialization

laserchicken:
Ravi reported "On the homepage, with the new header, the Shop link doesn't appear to be working".
This is because mobile-navigation element has `z-index: 10000` and before it's detached it covers `shop` and `sing in` buttons.

### DIFF
--- a/src/components/navigation/_mobile-navigation.scss
+++ b/src/components/navigation/_mobile-navigation.scss
@@ -22,6 +22,10 @@
     transition: transform ($animation-speed / 2), opacity ($animation-speed / 2);
   }
 
+  &--hidden {
+    z-index: -20;
+  }
+
   &--visible,
   &--visible &__list {
     opacity: 1;
--- a/src/components/navigation/_mobile-navigation.scss
+++ b/src/components/navigation/_mobile-navigation.scss
@@ -23,7 +23,7 @@
   }
 
   &--hidden {
-    z-index: -20;
+    z-index: z("below");
   }
 
   &--visible,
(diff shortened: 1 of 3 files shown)


jcreamer898:
Seems legit... :+1:

src/components/navigation/_mobile-navigation.scss
```diff
@@ -22,6 +22,10 @@
     transition: transform ($animation-speed / 2), opacity ($animation-speed / 2);
   }
 
+  &--hidden {
+    z-index: -20;
```
thomasthesecond: Why not just `-1`?

thomasthesecond:
@jcreamer898 Doesn't the mobile nav get removed from the DOM on the desktop view?

src/components/navigation/_mobile-navigation.scss
thomasthesecond: Also, we have a `z-index` map that you can use.

`z-index: z("below");` which outputs `z-index: -1;`

laserchicken:
@thomasthesecond it gets removed [here](https://github.com/lonelyplanet/rizzo-next/pull/408/files#diff-2cb3571c8b5ada02617297a2a71049c3R44) but before that happens it covers buttons (with cache cleared for few seconds); I'm just trying to fix what was reported as poor UX.

thomasthesecond:
Ok cool. I just wanted to make sure the JS wasn't broken because that was an issue previously.

PR #414, opened 2016-02-12
Namespace header component

### LINKED ISSUE #317: Namespace header component
We've already namespaced the search component for use alongside Rizzo and we should do the same for the header component. This also includes all sub-components of the header; logo, navigation, mobile navigation, sub-navigation, avatar and notification badge.

I also think we should consider renaming `.header` to `.global-header` in order to make it more specific. It will also allow us to use `.header` as a more generic header for things such as articles.

thomasthesecond:
Resolves #317

Related to https://github.com/lonelyplanet/destinations-next/pull/737

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -65,7 +65,7 @@ body {
       padding: 2.2rem 2rem 1.8rem;
       text-decoration: none;
       width: auto;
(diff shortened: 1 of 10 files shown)


jcreamer898:
:+1

PR #415, opened 2016-02-12
Remove top border on masthead text wrap

### LINKED ISSUE #395: Double border on destinations without images
See https://www.lonelyplanet.com/venezuela/barcelona

thomasthesecond:
It's not needed in any instance because there is a bottom border on the header.

`top: 0` can also be removed safely on narrow text wrap because `top: 0` is already defined on text wrap.

Fixes #395


jcreamer898:
:+1:

PR #416, opened 2016-02-15
Check for related articles first

thomasthesecond:
Fixes lonelyplanet/destinations-next#743.

Related to the sticky footer, there was a JS error when there are no related articles. This fix conditionally loads the sticky footer and other related code for infinite scrolling.

Also, fix a z-index bug on the sticky footer.


jcreamer898:
:+1:

PR #421, opened 2016-02-15
Change this.$newArticle

thomasthesecond:
`this.$newArticle` is already a jQuery object. Not sure if this will fix lonelyplanet/destinations-next#754.

### DIFF
--- a/src/components/article/article.hbs
+++ b/src/components/article/article.hbs
@@ -17,7 +17,7 @@
           data-targeting='{ "position": "header" }'>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


jcreamer898:
I'd be *surprised* if this fixes it TBH.

thomasthesecond:
I don't think it will. I've found where the issue is coming from, but can't reproduce it.

jcreamer898:
:+1:

jcreamer898:
Let's give this a whirl for a while and see what happens.

PR #422, opened 2016-02-15
Add CSS breakpoint values to JS

thomasthesecond:
It's not perfect, but this PR duplicates the breakpoint variables defined [here](https://github.com/lonelyplanet/rizzo-next/blob/tc-js-breakpoints/sass/_variables.scss) for use in JS with `matchMedia`.


jcreamer898:
:+1:

PR #423, opened 2016-02-16
Add timeout to fix articles with adblock

thomasthesecond:
The infinite scrolling experience for articles doesn't work when an ad blocker is present. This PR adds a timeout to the `adLoadedPromise` to allow infinite scrolling to work without ads.


thomasthesecond:
:+1: from @jcreamer898

PR #424, opened 2016-02-16
Recalculate amount to scroll for sticky footer

thomasthesecond:
Fixes lonelyplanet/destinations-next#746

The problem was that the amount to scroll was incorrect (too small) when an article has images that are enlarged. The initial amount to scroll is calculated before the images load and when enlarged, the images increase the amount to scroll. The solution was to recalculate the amount to scroll once during the scroll event.

Also, for the sticky footer component, a state variable `isHidden` was added to improve performance a bit; now it only checks once to `show` or `hide`.

Lastly, the debounce was increased from `10` to `100` to improve performance.


jcreamer898:
:+1:

PR #425, opened 2016-02-16
Create utility for string helpers

thomasthesecond:
Use regex to match _all_ spaces in `slugify` method. Update article component and ad manager to use slugify helper.


jcreamer898:
Nice work man. :+1:

PR #429, opened 2016-02-17
Simplify and fix underline-link mixin

thomasthesecond:
Use `px` instead of `rem` for underlined link mixin. It seems that somewhere along the line, `rem` stopped working. Changing values to `px`; this might be better anyway as we want absolute control over the underline size.

The mixin was simplified as well since we don't need different underline styles.

Fixes lonelyplanet/destinations-next#762


jcreamer898:
:+1:

PR #432, opened 2016-02-18
Improve article when JS is disabled

thomasthesecond:
There are a few things here that will improve the article experience when JS is disabled.

1. Hide wrappers for ads
2. Hide article footer; this area contains the publish date (JS formatted) and the social share component (requires JS)
3. Add the global footer back in; when JS is disabled and the footer is missing, the page looks broken; the footer can be detached in a JS enabled environment


thomasthesecond:
Relates to https://github.com/lonelyplanet/destinations-next/pull/772

jcreamer898:
:+1:

PR #433, opened 2016-02-18
Fix for article sponsored bottom ads

jcreamer898:
Closes lonelyplanet/destinations-next#773.
Move template.
Stop removing ad if not loaded.


thomasthesecond:
:+1:

PR #434, opened 2016-02-18
Stop removing masthead ad if not loaded

thomasthesecond:


jcreamer898:
That one might be ok to keep in there, it's this custom template one that's fudged up because of how it works...

thomasthesecond:
Ok, good to know.

PR #436, opened 2016-02-19
Remove resrc.it urls from interests component

mkunkel:
We no longer use resrc.it and this is handled in the backend via ImageFormatter.


thomasthesecond:
:+1:

PR #438, opened 2016-02-22
Refactor structure using ITCSS principles

thomasthesecond:
This PR moves some files around and includes a little refactoring. There's a bit more work to do, for example, combine all mixins into one single file, but I'll do that in another PR. I tried to do some of that work here, but the diffs got all weird, so for this PR, I'd like to keep it simple and just set up the structure.


jcreamer898:
Love love love love love.

PR #441, opened 2016-02-22
Don't need the check anymore

jcreamer898:
Don't need the `show_cards` anymore because we're checking in the view model.


thomasthesecond:
:+1:

PR #443, opened 2016-02-22
Get next article sooner

### LINKED ISSUE #392: Next article loading animation is not always visible on mobile devices
Most times the loading animation doesn't appear at the bottom of the article like it should.

Noticed on iPhone 5s.

thomasthesecond:
Once an article is 80% viewed, the next one will be fetched. Hopefully, this will help with ad impressions and viewability.

Also decreased the debounce to trigger the scroll event more often.

Fixes #392


jcreamer898:
Doooo it.

PR #445, opened 2016-02-22
Update and refactor Sass files

thomasthesecond:
- Create a separate partial for deprecated mixins
- Move mixins to one partial
- Move some objects out of base/page and into partials


jcreamer898:
:+1:

PR #446, opened 2016-02-23
Improve article, move ad to bottom

thomasthesecond:
- Remove AJAX call on first article
- Add error and tracking for 500 error
- Load ad at 80% threshold, load at bottom of article


thomasthesecond:
Related to https://github.com/lonelyplanet/destinations-next/pull/788

jcreamer898:
:+1:

PR #447, opened 2016-02-23
Prevent multiple ad loads for same article

thomasthesecond:
When loading ads at an 80% threshold (i.e., 80% of the article has been read), an ad would sometimes load more than once for an article when it was between 80% and 100%. Setting a `hasFetched` property to the `activeArticle` fixes this bug.


jcreamer898:
Let's definitely review the heck out of this on staging before pushing a prod build.

PR #450, opened 2016-02-25
Upgrading to Babel 6

jcreamer898:
Had to bump a few packages.
Had to use `default` for the `Injector` thing... more here http://jonathancreamer.com/webpack-code-splitting-with-es6-and-babel-6/.
Update webpack config.


thomasthesecond:
:+1:

PR #452, opened 2016-02-25
Update masthead text

thomasthesecond:
- Set a max width and center it (matches max-width on slideshow)
- Override the max width provided by the container mixin; fixes an IE bug


jcreamer898:
:+1:

PR #453, opened 2016-02-25
Introduce normalize.css; separate from reset.css

thomasthesecond:
This PR adds the full normalize.css and separates normalize styles from reset
styles. Previously, there were some normalize styles mixed in with reset styles,
but it'd be better to include the full normalize.css rather than parts of it
(for better cross-browser consistency) and separate it from reset (since they
are technically two different things).

Also this PR changes box-sizing on the universal selector to inherit as this is
a slightly better practice. See http://cl.ly/1w3T0E302y03

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -8,6 +8,7 @@ exclude:
 - 'docs/**'
 - 'node_modules/**'
 - 'tmp/**'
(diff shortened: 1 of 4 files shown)


jcreamer898:
:+1:

PR #455, opened 2016-02-26
Create mixin for slider buttons

thomasthesecond:


jcreamer898:
Did you double check that this works ok on `/best-in-us`?

thomasthesecond:
Good call. I'll look at that.

PR #456, opened 2016-02-26
Use trackEvent instead of calling utag.view on article scroll event

tkali:
Instead of calling `utag.view` directly, route virtual page view event (scroll-loading new article) through our analytics API.


jcreamer898:
:+1:

PR #457, opened 2016-02-26
Audit colors

thomasthesecond:
This is a first pass to get colors normalized and assigned to variables. Each step is described in the commit messages below.

For the next pass, I want to create a consistent naming convention for colors and possibly use a Sass map to store the variables.


jcreamer898:
Rock on!

PR #460, opened 2016-02-29
Create marker object and mixin

thomasthesecond:
This is the diamond with a number used on top places and experiences, sights, and map.


jcreamer898:
Awesome dude.

PR #463, opened 2016-02-29
Update non-flexbox layout for masthead

thomasthesecond:
Centers text in IE 9 and other browsers that don't support flexbox. The no-flexbox styles needed to be moved to different elements.

Fixes lonelyplanet/destinations-next#812


jcreamer898:
:+1:

PR #465, opened 2016-03-03
Update style for masthead ad

thomasthesecond:
Also, adjust slideshow styles so that it stays centered.

Closes lonelyplanet/destinations-next#817


jcreamer898:
:+1:

PR #470, opened 2016-03-22
Adding sass settings JSON file

jcreamer898:
We'll eventually import this in to the node-sass task as well.

Putting it here will allow us to import this file in any repository where we're using Radium as well.


thomasthesecond:
:+1:

PR #473, opened 2016-03-22
Fix cut off header in the footer

mkunkel:
The newsletter signup header was cut off. By inheriting the line height of the parent element, this no longer happens.

Resolves https://github.com/lonelyplanet/destinations-next/issues/825


thomasthesecond:
:+1:

PR #475, opened 2016-03-25
Route events through lp.analytics.api.trackEvent

tkali:
Instead of calling `utag.link` manually, make use of `analytics.js` and call `lp.analytics.api.trackEvent`.


jcreamer898:
:+1:

PR #483, opened 2016-04-12
Adding in new common core file

jcreamer898:
We should be able to utilize this common file for a lot of our apps.

Similar to app_core in old rizzo.


thomasthesecond:
:+1:

PR #489, opened 2016-04-18
Allow html encoded values in guidebook subtitle

mkunkel:
This is because the subtitle is going to include currency values from Scooter, which uses html encoded values.


thomasthesecond:
👍

nicrocs:
👍

PR #494, opened 2016-04-20
Linking Newsletter heading to newsletter page

nicrocs:
Previously there was no link to the newsletter page from "Newsletter Sign In". This fixes that.


thomasthesecond:
👍

PR #507, opened 2016-05-17
Am masthead add on

amccadams:
Makes modal available in masthead so that clickable links can be used in masthead.


jcreamer898:
👍

PR #512, opened 2016-05-23
have url check for data attribute before insertion window location

nicrocs:


nicrocs:
@jcreamer898 I think this would be the solution to what we were discussing.

thomasthesecond:
👍

PR #514, opened 2016-05-23
updates mobile modal so that it scrolls to top when bottom button is …

amccadams:
Updates mobile modal so that it scrolls to top when bottom button is clicked.


nicrocs:
Looks good! My stuff has already gone in, so you may need to fix any merge conflicts. if not, 👍

PR #526, opened 2016-06-06
Am national parks

amccadams:
Add CTA to alert and make ebook set HTML ready.

### DIFF
--- a/sass/settings/_colors.scss
+++ b/sass/settings/_colors.scss
@@ -23,7 +23,7 @@ $color-light-slate: #e4eaf8 !default;
 $active-background-color: #f4f5f8 !default;
 $color-light-blue: #bdcee3 !default;
 $color-icon-blue: #2095cf !default;
-$color-lightblue: #e9f2f8 !default;
+$color-paleblue: #e9f2f8 !default;
 
 
 // Grayscale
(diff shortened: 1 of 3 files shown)


sass/settings/_colors.scss
```diff
@@ -23,6 +23,8 @@ $color-light-slate: #e4eaf8 !default;
 $active-background-color: #f4f5f8 !default;
 $color-light-blue: #bdcee3 !default;
 $color-icon-blue: #2095cf !default;
+$color-lightblue: #e9f2f8 !default;
```
nicrocs: sinc there already is a light blue, maybe we should name this something else? pale-blue?

src/components/modal/ebook/ebook.hbs
```diff
@@ -11,7 +11,7 @@
     <h1>{{ebook.strapline}}</h1>
   </div>
   <div class="ebook__copy">
-    {{ebook.copy}}
+    {{{ebook.copy}}}
```
nicrocs: should we move this to marketing and have best in europe point there instead?

amccadams: I think so.  Can we wait until after we get National Parks taken care of?

nicrocs: yeah, works for me

nicrocs:
👍

PR #527, opened 2016-06-07
Ad position for article ad, style tweaks

jcreamer898:
A few small tweaks.

### DIFF
--- a/src/components/ads/ads.scss
+++ b/src/components/ads/ads.scss
@@ -102,7 +102,11 @@
 }
 
 .adunit--experiences {
-  padding-bottom: 20rem;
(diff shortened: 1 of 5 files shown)


src/components/ads/ads.scss
```diff
@@ -102,7 +102,11 @@
 }
 
 .adunit--experiences {
-  padding-bottom: 20rem;
+  padding-bottom: 10rem;
```
nicrocs: not sure what all this will affect, just making sure we don't have any unintended side effects here.

nicrocs:
as long as the change to padding in the adunit-experiences class is isolated, 👍

PR #531, opened 2016-06-28
Fix issue with overlay for shop

jcreamer898:
The `.overlay` class was overriding some things in Rizzo and covering the page on shop.

### DIFF
--- a/src/components/overlay/index.js
+++ b/src/components/overlay/index.js
@@ -10,7 +10,7 @@ class Overlay extends Component {
     this.options = options;
 
     this.$html = $("html");
-    this.$el.addClass("overlay");
+    this.$el.addClass("lp-overlay");
 
     this.events = {
       "click": "onClick",
@@ -45,7 +45,7 @@ class Overlay extends Component {
     getScrollbarWidth()
       .then((scrollWidth) => {
         setTimeout(() => {
-          this.$el.addClass("overlay--visible");
+          this.$el.addClass("lp-overlay--visible");
         }, 10);
 
         if (this.options.preventScroll) {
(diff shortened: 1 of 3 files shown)


src/components/overlay/index.js
```diff
@@ -10,7 +10,7 @@ class Overlay extends Component {
     this.options = options;
 
     this.$html = $("html");
-    this.$el.addClass("overlay");
+    this.$el.addClass("lp-overlay");
```
nicrocs: Is there a corresponding lp-overlay class that will give it the correct styling? I think this will break countries and cities on destinations pages.

src/components/overlay/index.js
```diff
@@ -45,7 +45,7 @@ class Overlay extends Component {
     getScrollbarWidth()
       .then((scrollWidth) => {
         setTimeout(() => {
-          this.$el.addClass("overlay--visible");
+          this.$el.addClass("lp-overlay--visible");
```
nicrocs: same thing here. I'm not seeing an lp-overlay--visible class to change the opacity

src/components/overlay/index.js
jcreamer898: I forgot to push the scss change. Lolz.

src/components/overlay/index.js
jcreamer898: fixed now

jcreamer898: good catch by the way, proves you're actually PRing me. :-D

nicrocs:
global CSS fun! 👍

jcreamer898:
indeed

PR #533, opened 2016-06-29
Fix relative links

jcreamer898:
The links on shop are breaking because they're relative.

### DIFF
--- a/lib/data/default.json
+++ b/lib/data/default.json
@@ -25,43 +25,43 @@
             "items": [
               {
                 "title": "Africa",
-                "slug": "africa"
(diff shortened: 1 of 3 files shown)


lib/data/default.json
```diff
@@ -25,43 +25,43 @@
             "items": [
               {
                 "title": "Africa",
-                "slug": "africa"
+                "slug": "https://www.lonelyplanet.com/africa"
```
nicrocs: small thing, but should these be called slugs now that they are the full urls?

nicrocs:
something to consider re: slugs/urls, otherwise 👍

PR #539, opened 2016-07-18
Get hotels component title from the view

mkunkel:
This will allow for promotions like the Rio takeover to alter the title of the hotels component.


nicrocs:
👍

PR #540, opened 2016-07-19
Change shop link if cart has items

thomasthesecond:
Issue detailed at https://github.com/lonelyplanet/rizzo/issues/1545


nicrocs:
👍

PR #541, opened 2016-07-22
Fix slug in masthead nav

thomasthesecond:
The slug contains a slash and for some reason, putting an extra slash in the `href` removes the slash instead of doubling it.


nicrocs:
👍

PR #548, opened 2016-08-16
bump patch version

laserchicken:


thomasthesecond:
I've bumped and published yesterday via https://github.com/lonelyplanet/rizzo-next/pull/547

PR #549, opened 2016-08-16
Use dataLayer.events to store adblock information

tkali:
Adblock checker loads before Google Analytics so in order to send the event, store it for later in `lp.analytics.dataLayer.events`.


jcreamer898:
👍

PR #551, opened 2016-08-23
use absolute path in items links

laserchicken:
request: https://github.com/lonelyplanet/search/issues/191
it won't work for apps deployed at domains different than https://www.lonelyplanet.com (like shop.lonelyplanet.com)


src/components/search/search_item/index.html.hbs
```diff
@@ -1,4 +1,4 @@
-<a class="lp-search-item__link" href="/{{slug}}" data-lpa-category="Destinations Next" data-lpa-action="Search Click" data-lpa-label="{{name}}">
+<a class="lp-search-item__link" href="https://www.lonelyplanet.com/{{slug}}" data-lpa-category="Destinations Next" data-lpa-action="Search Click" data-lpa-label="{{name}}">
```
nicrocs: any potential issues from having the https there? could it be on any domains that aren't https?

laserchicken: I don't expect any issues, those items links will point to waldorf, dn and hotels, all those applications are accessible over https. As far as I know we are in the process of transitioning all apps to https only, so I would not use http here.

robuye: correct, https should be available in all those apps.

nicrocs:
👍

Issue #554, opened 2016-09-01
Shop's lp logo should send user back to homepage


chrissysievers:
User Story
As a user you should be directed back to the homepage when clicking on the LP logo in Shop. 

Acceptance Criteria
- [x] LP logo in shop header should link directly to main homepage.

Other info:
- Due to rizzo-next using relative urls in links everywhere,
- See simliar issue recently closed https://github.com/lonelyplanet/search/issues/191

laserchicken:
fixed

PR #556, opened 2016-09-05
use absolute url for header lp logo link

laserchicken:
request: https://github.com/lonelyplanet/rizzo-next/issues/554

### DIFF
--- a/src/components/header/header.hbs
+++ b/src/components/header/header.hbs
@@ -4,7 +4,7 @@
   <div class="lp-global-header__container">
     <div class="lp-global-header__inner js-lp-global-header-inner">
 
-      <a class="lp-logo lp-logo--{{type}}" href="/">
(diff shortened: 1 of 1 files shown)


thomasthesecond:
Are we using https on the home page yet? I thought https was only being used for destinations next and new POIs at the moment. cc @jcreamer898

laserchicken:
@thomasthesecond I think we support ssl for www.lonelyplanet.com domain, but I just noticed there are problems with mixed content, maybe I should use http for now. more info: https://lonelyplanet.atlassian.net/wiki/display/OPS/SSL+Planning

laserchicken:
@thomasthesecond is this ok? Sebastian is waiting for it, I wanted to deploy it as soon as possible

thomasthesecond:
👍

PR #559, opened 2016-09-16
Add ad unit for new sponsor article

thomasthesecond:
https://github.com/lonelyplanet/destinations-next/issues/1038

### DIFF
--- a/src/components/articles/articles.hbs
+++ b/src/components/articles/articles.hbs
@@ -51,4 +51,11 @@
       data-targeting='{ "position": "article-list-last" }'>
     </div>
   </div>
+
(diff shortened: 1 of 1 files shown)


src/components/articles/articles.hbs
```diff
@@ -51,4 +51,11 @@
       data-targeting='{ "position": "article-list-last" }'>
     </div>
   </div>
+
+  <div class="article__sponsor-ad js-article-sponsor-ad">
```
nicrocs: I think they'll only need this one from now on, so you can remove the other one.

thomasthesecond: Ok, removed it.

PR #567, opened 2016-10-11
add adunit for badge

laserchicken:
request:

Add a new ad unit to support a “Best in” Badge for selected destinations. This ad unit should be created to perform like the unit that holds sponsorship logos on Destinations that have been sponsored. See https://www.lonelyplanet.com/ireland, and the small ad spot in the lower right hand side of the super zone. This is what we are needing for the lower LEFT hand side of the super zones.

*Requirements:*

- Create an add unit for the lower left corner of the superzone that acts just like the existing ad space  in the lower right corner, that’s populated with the “sponsored by” logo. (see screenshot below)

- This ad space should be available for any destination page to use (countries, cities, etc)

- The badge should be able to be populated to the destination via DFP.

-----------------------------------

New add unit will be targeted the same way `adunit--sponsor-logo-masthead` is (`window.lp.ads` config, per destination). Additionally there will be `"position": "masthead-best-in"` custom key targeting.

### DIFF
--- a/src/components/masthead/masthead.hbs
+++ b/src/components/masthead/masthead.hbs
@@ -68,7 +68,7 @@
     <div class="adunit adunit--best-in-masthead"
       id="best-in-badge-masthead"
       data-size-mapping="sponsor-logo"
-      data-targeting='{ "position": "masthead-best-in" }'>
+      data-targeting='{ "position": "masthead-best-in" }' data-dfp-options='{ "namespace": "LonelyPlanet.com/Editorial" }'>
     </div>
   </div>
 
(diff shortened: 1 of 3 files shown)


src/components/masthead/masthead.hbs
```diff
@@ -65,6 +65,11 @@
       data-size-mapping="sponsor-logo"
       data-targeting='{ "position": "masthead" }'>
     </div>
+    <div class="adunit adunit--best-in-masthead"
+      id="best-in-badge-masthead"
+      data-size-mapping="sponsor-logo"
+      data-targeting='{ "position": "masthead-best-in" }'>
+    </div>
```
nicrocs: In addition you will need to add a data-dfp-options attribute like so: https://github.com/lonelyplanet/destinations-next/blob/995853a815b64adb330029bda3e31b4c20440532/app/templates/layouts/application.html.hbs#L40
the name space will be "LonelyPlanet.com/Editorial/Best_in_Award_Logo"

PR #580, opened 2016-11-04
Updated styling on video poster button

harterc1:
Removed scaling of play button on hover.  The background color (black) still becomes more opaque as you hover.


nicrocs:
👍

PR #582, opened 2016-11-10
Style bug fixes for video player

harterc1:
- Fixed black screen when in video player fullscreen mode
- Fixed responsiveness of video player
- Added temporary 'video=true' query parameter to enable video poster button
- Added workaround for low-res brightcove video when high-res video is available


harterc1:
- Also, made it so the poster image button doesn't appear until the poster image has fully downloaded

PR #585, opened 2016-11-18
add header with normal settings to build process

nickfranciosi:
When importing rizzo-next components into projects using react we need a way to pull in the two different style headers that have already been built. 

This PR adds a separate config file that builds the header with the settings needed to accomplish this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 logs
 *.log
 
+#DS_Store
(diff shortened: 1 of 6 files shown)


jcreamer898:
Nice work dude.

nickfranciosi:
@jcreamer898 you good with me merging this in?

PR #586, opened 2016-11-21
Video poster button video opens "inline" instead of in an overlay

harterc1:
- Updated styling on Video Poster Button (Blue play icon design)
- Moved videojs specific CSS to it's own file
- Moved common logic to resize video player into Brightcove.js as it is used by both "Video Poster Button" and "Video Overlay"
- Brightcove player now resizes itself when it's initially loaded to ensure high res video content is loaded
- Brightcove player now automatically inserts LD-JSON (Schema.org) markup into page head tag when any video is loaded into the player.
- Added "ended" event emitted from Brightcove.js player.  Video Poster Button listens to this to know when the video ends and the poster image can be placed back on the page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "scsslint": "scss-lint . --config .scss-lint.yml",
     "prepublish": " rimraf dist && mkdir -p dist && node bin/rizzo build && mkdir -p tmp && cp -R src/* tmp && babel -d tmp tmp && cp -R tmp/* dist && rimraf tmp"
   },
-  "version": "0.11.88",
+  "version": "0.11.89",
   "pre-commit": [
     "scsslint",
     "lint"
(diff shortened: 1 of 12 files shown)


jcreamer898:
👍

thomasthesecond:
Does this update break backwards compatibility with apps that are already using the video overlay functionality? I would think we would want the video overlay to still work on apps/pages where it was intended to.

harterc1:
@thomasthesecond The video overlay was added to the site 2 weeks ago when the "video poster button" was implemented so at this point, that is the only component that was using it.  BUT I did make sure that the video overlay still works exactly the same as it did before I moved the code around before submitting this PR.   As far as I know, with this update, the video overlay is not used by anything, but it's still functional, and although I don't usually like leaving code around that isn't used, we may want it back at some point.

thomasthesecond:
@harterc1 Ok cool, I wasn't sure if the overlay was being used anywhere else. 👍

PR #590, opened 2016-12-06
Fix build issue

jcreamer898:
This highlights a bigger issue which is that the header build process desperately needs to be updated.

When the `node bin/rizzo build -c ../lib/data/header_normal.json` was added to `npm run build`, it was added after the original `node bin/rizzo build`. This was causing the `rizzo-next.js` file to only contain the constructors listed in... https://github.com/lonelyplanet/rizzo-next/blob/master/lib/data/header_normal.json.

The build process for the header was a bandaid for getting us a global nav and we need to pay down that technical debt.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "scripts": {
     "test": "karma start --single-run",
     "ci": "karma start",
-    "build": "rimraf dist && mkdir -p dist/header_normal && node bin/rizzo build && node bin/rizzo build -c ../lib/data/header_normal.json",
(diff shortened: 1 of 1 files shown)


robuye:
:+1: 

do we know why it happened? I think rizzo-next was not bumped in Rizzo?

jcreamer898:
I bet since we're using `^11.83` it's still installing the latest patch version...

robuye:
is it necessary to publish new version to npm too?

PR #593, opened 2016-12-14
get closest feature manually in case queryRenderedFeatures throws error

laserchicken:
request: https://trello.com/c/TkQTYjLX/121-map-on-london-destination-page-doesn-t-work-properly-you-can-t-click-on-the-icon-on-the-map-to-bring-up-the-name-of-the-place

Sometimes, for maps of big cities (London, Mexico City etc.), when there is a lot of markers in one place mapbox-gl-js' `queryRenderedFeatures` method throws `Caught: Unimplemented type: 7`. I implemented workaround for this case - getting closest feature/poi manually. `_getClosestFeatures` takes longitude and latitude of current cursor position, calculates equirectangular projection distance for all markers and returns the closest one or nothing if calculated minimal distance is too long.

### DIFF
--- a/src/components/map/mapbox/markerset.js
+++ b/src/components/map/mapbox/markerset.js
@@ -200,7 +200,7 @@ class MarkerSet extends Component {
         minIndex = i;
       }
     }
-    if (minDistance < 1) {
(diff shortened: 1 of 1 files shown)


katelynp43:
👍

PR #595, opened 2016-12-21
360 Video support for video-based components

harterc1:
- Videos are queried via Brightcove's PlayList API instead of simply attempting to load them via an embed code and watching for failures (The API still returns a 404 if unable to find a video -- just like the embed code would) 
http://docs.brightcove.com/en/video-cloud/playback-api/getting-started/api-overview.html#methods

- Once a video is found, if it has a "360" tag on it, a 360 player embed code is dynamically loaded into the page.  Otherwise, the default 2D player is loaded dynamically.

- Brightcove component won't dynamically load an embed if it sees an embed code already on the page (within the component's parent element).

- Video embed no longer assumes it has the same dimensions as it's corresponding poster image.  They size themselves independently now.

- If a video is determined to be a 360 video on a browser that doesn't support 360 video, video components don't render, so the user won't even notice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "scsslint": "scss-lint . --config .scss-lint.yml",
     "prepublish": "npm run build && mkdir -p tmp && cp -R src/* tmp && babel -d tmp tmp && cp -R tmp/* dist && rimraf tmp"
   },
-  "version": "0.11.95",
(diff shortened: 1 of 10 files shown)


jcreamer898:
Love it dude.

PR #608, opened 2017-01-21
Manually merged past 360 video work from branch ch-video

harterc1:


harterc1:
360 video is not on our radar currently.  This work is outdated.  Closing.

PR #611, opened 2017-02-01
Include restaurants hidden div with data source

Keqi:
Related: https://github.com/lonelyplanet/destinations-next/pull/1152

### DIFF
--- a/src/components/food_and_drink/food_and_drink.hbs
+++ b/src/components/food_and_drink/food_and_drink.hbs
@@ -3,6 +3,12 @@
 <article class="food-and-drink">
   <h2 class="food-and-drink__heading">{{title}}</h2>
 
+  {{#if food_and_drink.data_source }}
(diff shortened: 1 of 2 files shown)


src/components/food_and_drink/food_and_drink.hbs
```diff
@@ -3,6 +3,12 @@
 <article class="food-and-drink">
   <h2 class="food-and-drink__heading">{{title}}</h2>
 
+  {{#if food_and_drink.data_source }}
```
mkunkel: Why would there not be a `data_source`?
https://github.com/lonelyplanet/destinations-next/pull/1152/files#diff-b11b1812eb2faebf6ef917762f6083caR16 shows that there should be a value of either "es" or "op"

Keqi:
@mkunkel You're right. I have just copied it from love letter section where condition was needed.

PR #620, opened 2017-02-13
Masthead images hidden div

Keqi:


src/components/masthead/masthead.hbs
```diff
@@ -89,5 +90,3 @@
   {{/if}}
 
 </article>
-
```
mkunkel: I understand why one empty line was deleted, but we should have a single blank line at the end of a file.

Keqi: @mkunkel When I looked at the file view in Github, it seems to have one blank line at the end of file.

Keqi: That is weird. My text editor adds one blank line at the end too.

Keqi: I recreated that PR here: https://github.com/lonelyplanet/rizzo-next/pull/621. It does not remove last empty line.

PR #623, opened 2017-03-02
Add hidden div with data source to interests component

Keqi:


src/components/interests/interests.hbs
```diff
@@ -1,6 +1,7 @@
 {{#if show_cards}}
   <article class="interests">
     <h2 class="interests__heading">{{ interests.title }}</h2>
+    <div id="interests-{{data_source}}" style="display:none"></div>
```
mkunkel: Shouldn't this be set through CSS instead of inline? I don't know if we have a `hidden` class or not.

Keqi: @mkunkel You might be right however I would have to change all of components to be consistent. Currently they all have hidden div with inline css.

On the other hand these hidden divs will disappear once OP integration is complete.

PR #630, opened 2017-03-15
fix: Unhide the published date

jcreamer898:
By order of @swobba! 😛


swobba:
Great. Many thanks @jcreamer898 @chrissysievers ;)

PR #633, opened 2017-03-22
Add hidden div to top places component

Keqi:


src/components/top_places/top_places.hbs
```diff
@@ -1,6 +1,8 @@
 <section class="top_places{{#if centered}} top_places{{ centered }}{{/if}}">
   <button class="top_places__close icon-close-small" type="button">Close</button>
 
+  <div id="top_places-{{top_places_data_source}}" style="display:none"></div>
```
thomasthesecond: Would prefer this uses our utility class `is-hidden`. Does this get shown?

Keqi: Hey @thomasthesecond , 

This is something I talked about with @mkunkel. These divs are  in each one of these components so I would need to refactor them all. That would make sense but they are going to be removed once I am finished with Destinations-Next refactor.

Keqi: Is it okay for you to merge it that way?

thomasthesecond: Ok sounds good to me

thomasthesecond:
👍

PR #637, opened 2017-03-30
fix(navigation): Add video to global nav

jcreamer898:
This could get hairy if people add more channels, but not sure what else to do for now.


nickfranciosi:
@jcreamer898 can you review this also. Make sure I'm not missing anything

nickfranciosi:
@harterc1 I updated the urls here to match the text change, let me know if that effect anything on the op side

PR #642, opened 2017-04-20
Update article layout, add ad slots into article body

thomasthesecond:
- Updates top of article layout (author, header) to better match travel news
- Adds ad slots after every 6th paragraph in article body

lonelyplanet/articles#6


thomasthesecond:
lonelyplanet/destinations-next#1261

thomasthesecond:
I'm going to close this and split these two features into separate pull requests so that the ad feature can be wrapped in a feature flag.

PR #648, opened 2017-05-03
Update mobile navigation for profile

thomasthesecond:
I'm considering this a work in progress because these updates are mostly cosmetic. The styles have been updated to match the design and the necessary components have been build and added to the mobile navigation. This does not yet connect to the new dotcom-connect app and it still uses Luna as its provider.

lonelyplanet/dotcom-profile#2

image

image


thomasthesecond:
Removed the "WIP" flag. It seems we can launch this using Luna for authentication along with #649 for now.

PR #649, opened 2017-05-04
feat: Global modal

jcreamer898:
`dotcom-core` is now a part of Rizzo Next to render the global modal.
The code lives in `header_component`.


lib/data/default.json
```diff
@@ -145,7 +145,7 @@
         },
         {
           "title": "Sign In",
-          "slug": "//auth.lonelyplanet.com/users/sign_in"
+          "slug": "#login"
```
thomasthesecond: Is there a way we can keep the full URL and `preventDefault` if JS is loaded? We do this with the search button; it links to lonelyplanet.com/search but still opens the overlay. I'm asking because I think it's a nice way to let people still get to the sign in page if JS hasn't or doesn't load.

src/components/header/header_component.js
```diff
@@ -34,7 +35,37 @@ class Header extends Component {
     this.$mobileNotificationBadge = require("./mobile_notification_badge.hbs");
 
     this.appendMenuIcon();
+    this.buildGlobalComponents(options);
   }
+
+  buildGlobalComponents(options = {}) {
+    $(document).one("mouseenter", "a[href*='login']", (options) => this.lazyLoadGlobalComponents(options));
```
thomasthesecond: Curious why you chose to target the element like this `a[href*='login']` instead of using a class or and ID.

package.json
```diff
@@ -36,6 +36,7 @@
     "lint"
   ],
   "dependencies": {
+    "@lonelyplanet/dotcom-core": "^1.1.4",
```
thomasthesecond: Does this live in a repo somewhere?

thomasthesecond:
I think we can merge this and #648 and the same time.

lib/data/default.json
jcreamer898: Good idea!

src/components/header/header_component.js
jcreamer898: Because now you can open the modal from any link on any page by simply adding `#login` to it.

package.json
jcreamer898: https://github.com/lonelyplanet/dotcom-ui

src/components/header/header_component.js
thomasthesecond: Cool

package.json
thomasthesecond: Nice. I thought maybe that was it, thanks!

thomasthesecond:
https://github.com/lonelyplanet/dotcom-profile/issues/61

PR #662, opened 2017-05-18
Fix article ad, footer layout

thomasthesecond:
Removes the container styles from the article ad and footer
to align them with the article body.


thomasthesecond:
https://github.com/lonelyplanet/articles/issues/11

PR #663, opened 2017-05-19
Change ID for article ad slots, use feature flag

thomasthesecond:


thomasthesecond:
Helps with https://github.com/lonelyplanet/articles/issues/7

PR #680, opened 2017-06-13
Add toast to global header component

thomasthesecond:
Reads values from a browser cookie to display the toast.

The cookie `lpToast` should be set with JSON data (this example uses Rizzo next's `CookieUtil`):

```
this.cookieUtil.setCookie("lpToast", JSON.stringify({
  "message": "Message",
  "title": "Title",
  "type": "success"
}), 1);
```

* `message`: required
* `type`: required; one of “error”, “info”, “success” or “warning”
* `title`: optional

Duration is hard-coded as 3 seconds, but that value could be set in the cookie as well if certain toasts need more or less time on screen.

### DIFF
--- a/src/components/header/header_component.js
+++ b/src/components/header/header_component.js
@@ -1,4 +1,5 @@
 import { Component } from "../../core/bane";
+import CookieUtil from "../../core/cookie_util";
 import SearchComponent from "../search";
 import NavigationComponent from "../navigation";
 import NavigationState from "../navigation/navigation_state";
@@ -36,6 +37,8 @@ class Header extends Component {
 
     this.appendMenuIcon();
     this.buildGlobalComponents(options);
+
+    this.cookieUtil = new CookieUtil();
   }
 
   buildGlobalComponents(options = {}) {
@@ -64,25 +67,24 @@ class Header extends Component {
       toast.id = "lp-global-toast";
       document.body.appendChild(toast);
 
-      const toastTitle = sessionStorage.getItem("lp-toastTitle");
-      const toastMessage = sessionStorage.getItem("lp-toastMessage");
-      const toastType = sessionStorage.getItem("lp-toastType");
+      const toastData = this.cookieUtil.getCookie("lpToast");
       const toastDuration = 3000;
 
-      if (toastMessage && toastType) {
+      if (toastData) {
+        const data = JSON.parse(toastData);
+
         render({
           component: "GlobalToast",
           el: toast,
           props: {
-            title: toastTitle,
-            message: toastMessage,
-            type: toastType,
+            title: data.title,
+            message: data.message,
+            type: data.type,
             duration: toastDuration,
           },
         });
 
-        sessionStorage.removeItem("lp-toastMessage");
-        sessionStorage.removeItem("lp-toastType");
+        this.cookieUtil.removeCookie("lpToast");
 
         setTimeout(() => {
           document.body.removeChild(toast);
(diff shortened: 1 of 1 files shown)


thomasthesecond:
https://github.com/lonelyplanet/dotcom-profile/issues/60

src/components/header/header_component.js
```diff
@@ -60,14 +60,34 @@ class Header extends Component {
...
+          el: toast,
+          props: {
+            title: toastTitle,
+            message: toastMessage,
+            type: toastType,
+            duration: toastDuration,
+          },
+        });
+
+        sessionStorage.removeItem("lp-toastMessage");
```
toddkazakov: You should delete the title too

toddkazakov:
Since, accessing cross-domain local storage isn't possible, how should we display a notification after the user is successfully logged in? Isn't it better to use cookies instead?

thomasthesecond:
Yeah, we probably should use cookies. I didn't realize that you can't access session storage across domains, but it makes sense.

src/components/header/header_component.js
jcreamer898: You can store an entire Object in storage as well...

```js
sessionStorage.setItem("lpToastMessage", JSON.stringify({ "message": "foo", "title": "Foo title", type: "type" }));
const item = sessionStorage.getItem("lpToastMessage");

if (item) {
  const message = JSON.parse(item);
}
```

thomasthesecond:
@toddkazakov @jcreamer898 I've replaced session storage with a cookie and am using JSON for the data.

```
this.cookieUtil.setCookie("lpToast", JSON.stringify({
  "message": "You got a message?",
  "title": "Message title is here",
  "type": "success"
}), 1);
```

src/components/header/header_component.js
```diff
@@ -60,14 +63,33 @@ class Header extends Component {
...
+        const data = JSON.parse(toastData);
+
+        render({
+          component: "GlobalToast",
+          el: toast,
+          props: {
+            title: data.title,
+            message: data.message,
+            type: data.type,
+            duration: toastDuration,
```
nickfranciosi: @thomasthesecond should we add the `onClose` function through the props to also remove the cookie?

src/components/header/header_component.js
```diff
@@ -60,14 +63,33 @@ class Header extends Component {
...
+          },
+        });
+
+        this.cookieUtil.removeCookie("lpToast");
```
thomasthesecond: I remove the cookie here. Do you think we should or would need to remove it via `onClose`?

thomasthesecond: @nickfranciosi

nickfranciosi: I was thinking both. I know UX was wanting the `x` to appear with the toast. So if they don't dismiss the toast it will just clear based on your timeout

thomasthesecond: I'm passing `onClose` to the toast in dotcom-core

https://github.com/lonelyplanet/dotcom-ui/blob/master/packages/dotcom-core/src/components/toast/index.tsx#L79

But I'm not removing the cookie. I'm just changing the state so the toast disappears, and then after the `toastDuration`, the cookie is removed with the code above. I can remove the cookie in the `onClose` in dotcom-core

thomasthesecond: I think I can also do something like this:

```
props: {
  title: data.title,
  message: data.message,
  type: data.type,
  duration: toastDuration,
  onClose: () => { this.cookieUtil.removeCookie("lpToast"); },
},
```

nickfranciosi: oh ok cool, I just didnt' see the `x`

thomasthesecond:
I like the idea of passing an onClose prop and it works well 👍

thomasthesecond:
Fixed left and right margins

image

PR #681, opened 2017-06-15
Updated ImageGallery to support Brightcove videos

harterc1:
- Youtube videos are still supported
- Youtube and Brightcove videos are inserted via the `Video` component interface - so all methods in `ImageGallery` are now generic
- `Brightcove` player has been updated to insert it's own embed code instead of relying on it to already exist on the page (so removed the embed code from all templates as well)


harterc1:
These changes were merged into the `0.14.x` branch instead of `master` (for now)

PR #682, opened 2017-06-15
Video: Lower-third ad support; Brightcove support in ImageGallery

harterc1:


harterc1:
This was an accidental PR; Was supposed to merge into 0.14.x instead of master.  Closing.

PR #684, opened 2017-06-27
feat: Support dotcom-connect auth

jcreamer898:


src/components/navigation/navigation_component.js
```diff
@@ -192,6 +192,10 @@ class NavigationComponent extends Component {
     let $mobileNavigationHeader = this.$mobileNavigation.find(".js-mobile-navigation-header");
 
     if (!user.id) {
+      if (this.showNewLoginLink()) {
+        $(".navigation__link[href*='sign_in']", $(".lp-global-header__navigation"))
```
thomasthesecond: Should we use `js-*` classnames for these?

jcreamer898: It's hard with the global header because it's dynamic. We don't have great hooks in place.

thomasthesecond: Ok

PR #697, opened 2017-07-06
fix: change className for destinations navigation link.

shawncothran:
previously hardcoded to `--active`


thomasthesecond:
This change will be made anywhere the global header is required, not just Destinations Next. Is this what we want?

PR #708, opened 2017-07-11
Bump version to 0.19.2

Keqi:


thomasthesecond:
0.19.2 has been published. Just so you know, you don't need to manually change the version in package.json and create a pull request any more. Just let @jcreamer898 or myself know you need a release and we can do it automatically using standard-version, which also updates the changelog for us. 👍

PR #710, opened 2017-07-12
chore: bump backpack-ui through dotcom-core

nickfranciosi:


nickfranciosi:
no longer necessary due to pulling in global header directly from dotcom-core

PR #718, opened 2017-07-17
feat: add beta banner for profile launch

nickfranciosi:
adds beta banner with feedback and opt out links based on the profile variant cookie

### DIFF
--- a/sass/settings/_colors.scss
+++ b/sass/settings/_colors.scss
@@ -12,6 +12,7 @@ $color-orange: #ffc83f !default;
 $color-yellow: #feef6d !default;
 
 $active-color: #0fdebd !default;
+$beta-banner-color: #1bca8f !default;
 $color-red: #dc221a !default;
 $color-gray: #e4e4e4 !default;
 $color-light-gray-blue: #f0f6f9 !default;
--- a/src/components/beta_banner/beta_banner.hbs
+++ b/src/components/beta_banner/beta_banner.hbs
@@ -1,0 +1,13 @@
+<div class="lp-beta-banner">
+  <div class="lp-beta-banner__container">
+      <div class="lp-beta-banner--left">
+        <p>Welcome to our new beta experience! Help us test new functionality.</p>
+      </div>
+      <div class="lp-beta-banner--right">
+          <ul class="lp-beta-banner--action-list">
+            <li><a id="feedback-beta" href="#">Feedback</a></li>
+            <li><a id="exit-beta" href="#">Exit Beta</a></li>
+          </ul>
+      </div>
+  </div>
+</div>
--- a/src/components/beta_banner/index.js
+++ b/src/components/beta_banner/index.js
@@ -1,0 +1,1 @@
+import "./_beta_banner.scss";
--- a/src/components/header/header_component.js
+++ b/src/components/header/header_component.js
@@ -2,8 +2,10 @@ import { Component } from "../../core/bane";
 import SearchComponent from "../search";
 import NavigationComponent from "../navigation";
 import NavigationState from "../navigation/navigation_state";
+import CookieUtil from "../../core/cookie_util";
 import $ from "jquery";
 import debounce from "lodash/debounce";
+import BetaBannerComponent from "../beta_banner/beta_banner_component";
 
 /**
  * The page header which contains both search and navigation.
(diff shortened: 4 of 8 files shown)


sass/settings/_colors.scss
```diff
@@ -12,6 +12,7 @@ $color-orange: #ffc83f !default;
 $color-yellow: #feef6d !default;
 
 $active-color: #0fdebd !default;
+$beta-banner-color: #1bca8f !default;
```
thomasthesecond: In Backpack, we have `accentGreen` as #16c98d, which is nearly identical, but I wonder if we should use it instead?

thomasthesecond: Maybe it doesn't matter here

src/components/beta_banner/beta_banner.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="lp-beta-banner">
+  <div class="lp-beta-banner__container">
+      <div class="lp-beta-banner--left">
+        <p>Welcome to our new beta experience! Help us test new functionality. {{test}}</p>
```
thomasthesecond: Should this `{{test}}` be in here?

nickfranciosi: nope - good catch!

src/components/header/header_component.js
```diff
@@ -64,6 +70,46 @@ class Header extends Component {
...
+      e.preventDefault();
+      this.leaveBeta();
+    });
+  }
+
+  leaveBeta() {
+    this.cookieUtil.removeCookie("_v"); // Remove Variant
+    window.location = "//connect.lonelyplanet.com/users/sign_out"; // Sign Out
+  }
+
```
thomasthesecond: I wonder if all of this logic for the banner should be extracted out into beta_banner/index.js and then imported into the header so it's easier to remove when we're done with it?

I think you would use it like `this.betaBanner = new BetaBanner();`

nickfranciosi:
@thomasthesecond updated

src/components/header/header_component.js
```diff
@@ -2,8 +2,10 @@ import { Component } from "../../core/bane";
 import SearchComponent from "../search";
 import NavigationComponent from "../navigation";
 import NavigationState from "../navigation/navigation_state";
+import CookieUtil from "../../core/cookie_util";
 import $ from "jquery";
 import debounce from "lodash/debounce";
+import BetaBannerComponent from "../beta_banner/beta_banner_component";
```
thomasthesecond: I think you can import this into beta_banner/index.js and then just ``mport BetaBannerComponent from "../beta_banner/";

Like we do with https://github.com/lonelyplanet/rizzo-next/blob/master/src/components/footer/index.js

src/components/header/index.js
```diff
@@ -2,6 +2,7 @@ import HeaderComponent from "./header_component";
 
 import "./_header.scss";
 import "../logo/logo.scss";
+import "../beta_banner/_beta_banner.scss";
```
thomasthesecond: I _think_ you can remove this since you're importing into beta_banner/index.js above

nickfranciosi: I am not importing the scss anywhere else so I think it does need to be there, I can import someplace else though

thomasthesecond: You have it here:

https://github.com/lonelyplanet/rizzo-next/pull/718/files#diff-72ac1f525538159e260819c707524fef

But I don't think you're importing that anywhere. But if you look at my comment above and you import the SCSS and beta_banner_component into index.js, then you can just import index.js into header_component.js.

Hope that makes sense

src/components/beta_banner/index.js
```diff
@@ -0,0 +1 @@
+import "./_beta_banner.scss";
```
thomasthesecond: If you make this:

```
import BetaBanner from "./beta_banner_component";

require("./_beta_banner.scss");

export default BetaBanner;
```

Then in header_component, you can:

```
import BetaBanner from "../beta_banner";
```

and delete `import "../beta_banner/_beta_banner.scss";` from header/index.js

nickfranciosi: this is actually not working. I don't really see an issue with the banner css being included with the header. Just the same as the navigation , notification and logo scss are being included with the header. This is just another piece of the header.

PR #731, opened 2017-07-20
Mk profile link fix

laserchicken:


src/components/login/user.js
```diff
@@ -24,11 +24,11 @@ export default class User {
...
     this.connect = connect;
     this.profileLink = connect ?
-      `//www.lonelyplanet.com/profile/${id}/edit` :
+      `https://www.lonelyplanet.com/profile/${id}/edit` :
       `//www.lonelyplanet.com/thorntree/profiles/${profileSlug}`;
 
     this.publicProfileLink = connect ?
-      `/profile/${username}` :
+      `https://www.lonelyplanet.com/profile/${username}` :
       `//www.lonelyplanet.com/thorntree/profiles/${profileSlug}`;
```
thomasthesecond: Can we add `https` to the thorn tree URL also?

PR #737, opened 2017-08-02
fix: banner styles for 1%

nickfranciosi:
removes action buttons
updates text
resolves issue where banner could get added twice

### DIFF
--- a/src/components/beta_banner/_beta_banner.scss
+++ b/src/components/beta_banner/_beta_banner.scss
@@ -3,51 +3,16 @@
 .lp-beta-banner {
   background: $beta-banner-color;
   color: $color-white;
-  font-size: 1.6rem;
-  padding-bottom: 1.6rem;
-  padding-top: 1.6rem;
+  font-size: 1.4rem;
+  padding-bottom: 1.4rem;
+  padding-top: 1.4rem;
   width: 100%;
 
   @media (max-width: $max-720) {
-    font-size: 1.2rem;
-  }
-
-  &--left {
-    float: left;
-    max-width: 50%;
-  }
-
-  &--right {
-    position: absolute;
-    right: 0;
-  }
-
-  &--action-list li {
-    display: inline;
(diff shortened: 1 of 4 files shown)


src/components/beta_banner/_beta_banner.scss
```diff
@@ -3,51 +3,16 @@
 .lp-beta-banner {
   background: $beta-banner-color;
   color: $color-white;
-  font-size: 1.6rem;
-  padding-bottom: 1.6rem;
-  padding-top: 1.6rem;
+  font-size: 1.4rem;
+  padding-bottom: 14px;
+  padding-top: 14px;
```
thomasthesecond: Should the padding values be rem?

src/components/beta_banner/_beta_banner.scss
```diff
@@ -3,51 +3,16 @@
...
-    }
-
-    a:focus {
-      outline: 1px dotted rgba($color-white, .4);
-      outline-offset: 2px;
-    }
+    font-size: 1.1rem;
+    letter-spacing: -.65px;
+    padding-bottom: 8px;
+    padding-top: 8px;
```
thomasthesecond: Should the padding values be rem here too?

nickfranciosi: @thomasthesecond I have no idea, I feel like I hear something different every time I change the css in this project

thomasthesecond: Ok..it's probably fine as pixels 👍

nickfranciosi:
@thomasthesecond is that something new? I never added that

thomasthesecond:
No it's always been that way for global header. I added those styles a couple of weeks ago when @laserchicken noticed the problem.

Solution would be to use px in styles here, but we would have to remove those override styles.

thomasthesecond:
I can update those styles if you want

nickfranciosi:
I will update. Thanks

PR #745, opened 2017-08-09
Updated Brightcove videos to query playlists/videos through Open Planet

harterc1:
This should clean up a lot of the console error clutter:
- No longer uses Brightcove's API -- No more 404 errors
- Added error handling around videojs.play() -- No more "play interrupted" errors
- Video embed isn't created unless we have a video to load -- No more "videojs-contrib-ad" errors

### DIFF
--- a/src/components/video/brightcove.js
+++ b/src/components/video/brightcove.js
@@ -28,7 +28,12 @@ class Brightcove extends VideoPlayer {
   play() {
     super.play();
     this.autoplay = true;
-    this.player.play();
+    const promise = this.player.play();
+
+    // Catch any errors thrown within play promise (only applicable on some browsers)
+    if (promise) {
+      promise.catch(reason => console.log('VIDEOJS:', reason)).then(() => {});
+    }
   }
 
   pause() {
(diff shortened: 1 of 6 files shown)


src/components/video/brightcove.js
```diff
@@ -28,7 +28,12 @@ class Brightcove extends VideoPlayer {
   play() {
     super.play();
     this.autoplay = true;
-    this.player.play();
+    const promise = this.player.play();
+
+    // Catch any errors thrown within play promise (only applicable on some browsers)
+    if (promise) {
+      promise.catch(reason => console.log("VIDEOJS:", reason)).then(() => {});
```
thomasthesecond: Should this be in some kind of dev-only mode?

src/components/video/brightcove.js
```diff
@@ -196,42 +202,50 @@ class Brightcove extends VideoPlayer {
...
           this.loadNextVideo();
           resolve(true);
-        } else {
-          this.player.catalog.getVideo(query, (error, video) => {
-            if (!error) {
-              this.videos = [video];
+        }
+        else {
+          $.ajax({
+            url: "https://www.lonelyplanet.com/video/api/video.json?reference_id=" + query
```
thomasthesecond: Maybe we can create a constant for the API url since we use it a couple times.

`const apiUrl = "https://www.lonelyplanet.com/video/api/";`

src/components/video/brightcove.js
harterc1: It's something that is creating errors in the browser console that i don't want to be there, but the errors we're seeing don't seem to actually impact the user experience and clutter the console and trackjs.   So, this is basically just hiding it in all environments.

thomasthesecond: 👍

src/components/video/brightcove.js
harterc1: done.

PR #752, opened 2017-08-24
feat: add user id to dataLayer globally

nickfranciosi:
instead of setting user id to data layer app by app, this pr will set the id globally when user logs in

resolves https://github.com/lonelyplanet/dotcom-profile/issues/166

### DIFF
--- a/src/components/login/login_manager.js
+++ b/src/components/login/login_manager.js
@@ -40,6 +40,23 @@ export default class LoginManager {
       this.statusFetched(user);
     });
   }
+
+   /**
+   * Sets user id to dataLayer for analytics
+   * @return null
+   */
+  setDataLayer(id) {
+    if(
+      window &&
+      window.lp &&
+      window.lp.analytics &&
+      window.lp.analytics.dataLayer &&
+      window.lp.analytics.dataLayer.length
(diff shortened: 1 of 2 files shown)


src/components/login/login_manager.js
```diff
@@ -40,6 +40,23 @@ export default class LoginManager {
...
+   * Sets user id to dataLayer for analytics
+   * @return null
+   */
+  setDataLayer(id) {
+    if(
+      window &&
+      window.lp &&
+      window.lp.analytics &&
+      window.lp.analytics.dataLayer &&
+      window.lp.analytics.dataLayer.length
```
thomasthesecond: Very thorough

nickfranciosi: yeah... I hate having to do that

thomasthesecond: I know it's necessary

egdelwonk:
Eventually we could use the optional chaining operator. 

i.e. `window?.lp?.analytics?.dataLayer?.length`


Looks like there is a babel transform for it. https://github.com/babel/babel/tree/7.0/packages/babel-plugin-transform-optional-chaining

PR #772, opened 2017-10-03
Updated Video components for Best in Travel

harterc1:
1. Added `File` video type for general file video sources (embeds a simple `video` tag).
2. Add the following parameters to rizzo's video component:

    `playerName` - (optional) "default", "bestintravel", "destinations", or "background".  This determines
          which player from the respective "type" to insert into the page. ("brightcove" only).

    `poster` - (optional) URL to poster image to display before video begins
          playing ("file" only)

    `seo` - (optional) Whether to render SEO markup for the video ("brightcove" only)

    `controls` - (optional) Whether to include HTML5 video controls on the player or
          not ("brightcove" and "file" only)

    `popout` - (optional) Whether the player should follow the user when it scrolls
          out of view ("brightcove" only) -- Only works if the video embed does not
          exist on the page prior to instantiation.

    `cover` - (optional) Whether to cover the entire parent element with the
          video -- acts similar to background-size css. ("file" and "brightcove" only).

    `muted` - (optional) Whether the video should be muted or not ("file" and "brightcove" only).

    `playWhenInView` - (optional) Whether to play the video automatically once the
          player enters the viewport.  This will only trigger once during the lifetime of
          the player instance. ("brightcove" only).

Usage in masthead of BIT page:
screen shot 2017-10-03 at 10 26 46 am

Popout player (stays at 16:9 -- the video in the screenshot isn't 16:9 but it's just a test video):
screen shot 2017-10-03 at 10 27 08 am

Popout player (hover):
screen shot 2017-10-03 at 10 27 15 am

### DIFF
--- a/src/components/video/file.js
+++ b/src/components/video/file.js
@@ -11,12 +11,20 @@ class File extends VideoPlayer {
     if (this.videoId) {
       let html = "<video preload='auto' src='" + this.videoId + "' ";
 
+      if (this.cover) {
+        html += "class='video__cover' ";
+      }
+
       if (this.controls) {
-        html += " controls ";
+        html += "controls ";
+      }
+
+      if (this.muted) {
+        html += "muted ";
       }
 
       if (this.autoplay) {
-        html += " autoplay ";
+        html += "autoplay ";
       }
 
       if (this.poster) {
@@ -27,7 +35,13 @@ class File extends VideoPlayer {
 
       this.$el.html(html);
 
+      if (this.cover) {
+        this.$el.addClass("video__cover--container");
+      }
+
       this.player = this.el.getElementsByTagName("video")[0];
+      this.player.on("playing", () => { this.trigger("started"); });
+      this.player.on("ended", () => { this.trigger("ended"); });
     }
     super.setup();
   }
@@ -35,7 +49,10 @@ class File extends VideoPlayer {
   play() {
     super.play();
     if (this.player) {
-      this.player.play();
+      const promise = this.player.play();
+      if (promise) {
+        promise.catch(reason => console.log("VIDEOJS:", reason)).then(() => {});
+      }
     }
   }
 
--- a/src/components/video/file.js
+++ b/src/components/video/file.js
@@ -1,0 +1,55 @@
+import VideoPlayer from "./video_player";
+
+class File extends VideoPlayer {
+
+  initialize(options) {
+    this.player = null;
+    super.initialize(options);
+  }
+
+  setup() {
+    if (this.videoId) {
+      let html = "<video preload='auto' src='" + this.videoId + "' ";
+
+      if (this.controls) {
+        html += " controls ";
+      }
+
+      if (this.autoplay) {
+        html += " autoplay ";
+      }
+
+      if (this.poster) {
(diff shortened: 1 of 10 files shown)


src/components/video/index.js
```diff
@@ -75,6 +136,41 @@ class Video {
...
+      if (!youtubeId && !brightcoveId) {
+        return "file";
+      }
+    }
+
+    return type;
+  }
+
+  static cleanVideoId(videoId) {
+    if (videoId && videoId.toLowerCase().startsWith("http")) {
```
jcreamer898: Not sure if we have a polyfill for `startsWith` and it's not in IE. http://d.pr/i/6kJUL9, just double check that it still works and if not we'll have to pull in a polyfill for it.

src/components/video/video_player.js
```diff
@@ -14,6 +36,17 @@ class VideoPlayer extends Component {
...
 
   /**
+   * Use this to change any properties of the player after initialization.
+   * There is no default logic for this method as changing some properties may
+   * involve complex operations rather than simply changing the properties
+   * on the instance.
+   * Warning: This is implemented as needed and currently only covers "autoplay"
+   * and "videoId" on Brightcove players.
+   */
+  set(options) {
```
jcreamer898: This is empty?

src/components/video/index.js
harterc1: The search logic uses `startsWith` and i just went and tested on our live site to see if it works on IE and we're good there -- must have a polyfill someplace.

src/components/video/video_player.js
harterc1: This is only really implemented on the `Brightcove` player, but I put it here in the base class to show that it is intended to be a "public" method that all players support.   The reason it's blank is because (at least for the Brightcove player), the logic isn't as simple as `this.whatever = options.whatever`.  There is additional logic that runs when changing player properties.   So, i figure it can be implemented "as needed".

src/components/video/file.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    super.setup();
+  }
+
+  play() {
+    super.play();
+    if (this.player) {
+      const promise = this.player.play();
+      if (promise) {
+        promise.catch(reason => console.log("VIDEOJS:", reason)).then(() => {});
```
jcreamer898: Maybe we should do `trackJs.console.log`?

jcreamer898: At least I think that's what it is...

src/components/video/index.js
```diff
@@ -1,21 +1,89 @@
...
+import File from "./file";
 
 require("./_video.scss");
 
 let players = new Map();
 players.set("brightcove", Brightcove);
 players.set("youtube", Youtube);
+players.set("file", File)
 
+/*
```
jcreamer898: There's actually a known format for doing this called `jsdoc`... http://usejsdoc.org/

Essentially it'd be...

```js
/**
*  Video - an interface for ....
*  @class
*  @example 
*   let player = null;
*  @param {Element} element
*  @param {"brightcove" | "youtube" | "file"} type This determines which type of
*  @param {string} playerName
*
*/
```

PR #788, opened 2017-10-25
Updated video overlays for better quality of life on mobile

harterc1:


src/components/video/brightcove.js
```diff
@@ -194,7 +195,12 @@ class Brightcove extends VideoPlayer {
     }
     else if (!this.videoEl) {
       // Insert brightcove player html
-      let html = "<div class='video__popout'>";
+      let html = "<div class='video__popout ";
```
nickfranciosi: @harterc1 Instead of concatanating all of these strings you should probably be using template literals 

`<div class='video__popout ${MobileUtil.isMobile() && "video__popout-mobile" } />

harterc1: cool i can do that

PR #810, opened 2018-02-13
fix: show subnav for full length of destination page

nickfranciosi:
since we started adding sections to the destination pages, the subnav no longer stays visible by the time you get half way down the page

this removes the check for the bottom boundary to hide the subnav and leaves it visible the entire time scrolling

fix https://github.com/lonelyplanet/destinations-next/issues/1503

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -160,9 +160,31 @@
         "inversify": "4.3.0",
         "isomorphic-fetch": "2.2.1",
         "js-cookie": "2.1.4",
+        "radium": "0.18.4",
         "react": "15.6.1",
         "react-side-effect": "1.1.3",
         "reflect-metadata": "0.1.10"
+      },
+      "dependencies": {
+        "radium": {
+          "version": "0.18.4",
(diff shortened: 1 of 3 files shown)


thomasthesecond:
If this fix is for longer pages, it still seems like we might need the bottom boundary on shorter pages? From the code I'm guessing `this.contentHeight` is the amount that can be scrolled? Would we need to recalculate that somehow?

Also, in the issue you mention, the link to the flights widget is missing from the nav. It should be between books and activities. I wonder if the missing link in the nav is causing the nav scrolling to mess up?

thomasthesecond:
When we add a new section to a page, I thought it was a given that the corresponding link would be added to the sub-nav. The nav link highlights as you pass thru the section. Seems odd not to have it.

I didn't think we were hiding the sub-nav. I thought we were sticking/unsticking past a certain point. It would unstick at the bottom of the page (`isBottom`) so the nav didn't stick on top of the global footer. If this isn't an issue then I guess it's fine.

PR #817, opened 2018-03-14
Mk fix build

laserchicken:


laserchicken:
It seems newer npm do not install peer dependencies automatically anymore. Build was failing because react-dom (peer dependency) wasn't installed when tests were running. @lburdock @jcreamer898

PR #840, opened 2018-09-10
Change footer link for travel companions

nidalaa:


nidalaa:
Hi @thomasthesecond 👋 can we release new version, so I can reuse it in old rizzo? 🙏

thomasthesecond:
@nidalaa 0.35.5 is published.

PR #842, opened 2018-09-19
fix: reallow basic text search from nav

vacas5:
The search bar overlay was at one point working but then deactivated. My
best understanding was that it was the result of autocomplete not
working. Reimplementing without autocomplete to allow a user to make a
search query from anywhere on the site without having to drill down into
the search page first.

### DIFF
--- a/spec/unit/components/header.spec.js
+++ b/spec/unit/components/header.spec.js
@@ -1,7 +1,13 @@
 import $ from "jquery";
 let Injector = require("inject-loader!../../../src/components/header/header_component");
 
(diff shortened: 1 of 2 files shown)


nidalaa:
@vacas5 🤔 what is the end result? Is it going to show result list in drop down menu? Or just let user type in input and go to search page only after hitting `Enter` ?

vacas5:
yes, the user gets a text input in an overlay and then it redirects them to the search page on enter (or form submission on mobile). It's actually pretty clean, not sure why it was deprecated.

PR #848, opened 2018-11-15
Add tags section at the top of the article body

Keqi:
Related: https://github.com/lonelyplanet/destinations-next/pull/1631

Screenshots:

tags-1
tags-2

### DIFF
--- a/src/components/article_body/article_body.hbs
+++ b/src/components/article_body/article_body.hbs
@@ -1,7 +1,7 @@
 <div class="article-body has-legacy-code js-article-body" data-post-date="{{ article.post_date }}">
 
   <div class="article-tags">
-    {{#each article.tags}}
+    {{#each article.tags_for_ui}}
       <a href="https://www.lonelyplanet.com/search?topic_id={{this.id}}" target="_blank">
         <button class="tag">{{this.label}}</button>
       </a>
--- a/src/components/article_body/article_body.hbs
+++ b/src/components/article_body/article_body.hbs
@@ -2,7 +2,7 @@
 
   <div class="article-tags">
     {{#each article.tags}}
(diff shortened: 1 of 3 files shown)


src/components/article_body/article_body.hbs
```diff
@@ -1,4 +1,13 @@
 <div class="article-body has-legacy-code js-article-body" data-post-date="{{ article.post_date }}">
+
+  <div class="article-tags">
+    {{#each article.tags}}
+      <a href="https://www.lonelyplanet.com/search?tag={{this.id}}", target="_blank">
```
toddkazakov: Which one do we agree to use `tag` or `topic_id`?

vacas5: looks like @toddkazakov 's pr in dotcom web uses `topic_id` here. is there a reason we switched from tag? will it be tag in OP?

toddkazakov: There are 4 different types of tags in OP - legacy tags (user interests), categories, entities and topics. In the beginning we will surface only `topics`. @Keqi the query param that needs to be used is `topic_id`. As I cannot get this working locally, could you please discuss the UI changes with @bpotterbaum and push to QA for testing?

Keqi: @toddkazakov @vacas5 I renamed param name to topic_id and attached some screenshots to show you how it looks like. Do you think we can approve it? I would like to show it to Brad in Staging/Prod, there is no harm in deploying it because it is hidden behind a feature flag.

vacas5: there is an extraneous comma after the `href` tag on this line. Unless it's a rails thing I don't understand

Keqi: @vacas5 You're right. Sorry about that. Fixed now :)

PR #851, opened 2018-12-14
chore: excise lp-analytics

johnwquarles:
`rizzo` imports `rizzo-next`, but doesn't have access to all of its submodules, I suppose.

Since its runs across `lp-analytics` in the process of building out the head js & footer, it tries to import the module, but can't find it.

I think it could probably be made to work by importing `lp-analytics` via bower/require-js in `rizzo`, but I tbh don't know enough about how those differing module systems work to feel confident that I could resolve the issue quickly enough.

This is the hack solution-- do the same thing that `lp-analytics` is supposed to do for us.

(using `window.dataLayer` will work-- `lp-analytics` will be invoked through other means on the host page, and `window.dataLayer` will either alias to `window.lp.analytics.dataLayer` or it will itself be the source of truth for gtm content).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -234,17 +234,6 @@
         }
       }
     },
-    "@lonelyplanet/lp-analytics": {
-      "version": "2.0.9",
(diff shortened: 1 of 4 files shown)


sunderhaus:
This looks good to me.

PR #855, opened 2019-01-22
fix: adds conditional check on adunit in _type

Courey:
When adunit is undefined this error is thrown:
https://my.trackjs.com/errors/filtered?hash=9db7eab01247a509
Steps to repeat in production:
1. make sure adblocker is off
2. open console
3. visit a destination page
this is what I am getting in the console
screen shot 2019-01-22 at 2 04 30 pm

### DIFF
--- a/src/core/ads/ad_manager.js
+++ b/src/core/ads/ad_manager.js
@@ -96,7 +96,7 @@ export default class AdManager {
     if ($adunit && $adunit[0] && "id" in $adunit[0]) {
       return `${$adunit.data("sizeMapping")}-${$adunit[0].id}-${$adunit.data("adType") || "default"}`;
     } else {
-      return `${"default"}`;
+      return "default";
     }
   }
 
--- a/src/core/ads/ad_manager.js
+++ b/src/core/ads/ad_manager.js
@@ -93,7 +93,9 @@ export default class AdManager {
   }
 
   _track($adunit) {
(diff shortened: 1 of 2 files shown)


src/core/ads/ad_manager.js
```diff
@@ -93,7 +93,9 @@ export default class AdManager {
   }
 
   _track($adunit) {
-    return `${$adunit.data("sizeMapping")}-${$adunit[0].id}-${$adunit.data("adType") || "default"}`;
+    if ($adunit && $adunit.length) {
+      return `${$adunit.data("sizeMapping")}-${$adunit[0].id}-${$adunit.data("adType") || "default"}`;
```
vacas5: will still throw if `adunit[0]` is not an object. Might want to try something like lodash `get` function if that's a dependency, otherwise we should check if `id` is defined on `adunit[0]`

vacas5: the issue still remains that this will throw if `$adunit[0]` is a type other than an object but I'm not sure that's possible. But, I think you've solved the underlying issue of `$adunit[0]` coming back undefined.

Courey:
I haven't seen any errors for anything except when it's undefined. Should I work on preventing all errors or focus on fixing the one we know we've gotten?

vacas5:
I'd say just fix the one you can account for

Courey:
ok, this version fixes that one.

PR #856, opened 2019-01-25
fix: replaces a deprecated function to prevent error

Courey:
getDOMNode is no longer supported and calling it causes an uncaught error

### DIFF
--- a/src/components/map/views/sidebar-details.jsx
+++ b/src/components/map/views/sidebar-details.jsx
@@ -3,11 +3,16 @@ import MapActions from "../actions";
 import $clamp from "clamp-js/clamp.js";
 
 export default class SidebarDetailsView extends React.Component{
-  componentDidMount() {
-    let el = this.refs.poiTitle.getDOMNode();
+  constructor(props) {
+    super(props);
 
-    $clamp(el, { clamp: 2 });
+    this.setPoitTitleRef = el => this.poiTitle = el;
+
   }
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/components/map/views/sidebar-details.jsx
```diff
@@ -3,11 +3,16 @@ import MapActions from "../actions";
 import $clamp from "clamp-js/clamp.js";
 
 export default class SidebarDetailsView extends React.Component{
-  componentDidMount() {
-    let el = this.refs.poiTitle.getDOMNode();
+  constructor(props) {
+    super(props);
+
+    this.setPoitTitleRef = el => this.poiTitle = el;
```
vacas5: check and see if you're able to create a ref like this: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

creating refs has changed a bit as React has evolved, so depending on which version rizzo-next uses, it may be impossible.

Courey: this repo has "react": "^15.0.0 || ^16.0.0", and that was introduced in 16.3.

vacas5: when I run `npm install` it looks like v16.3.2 of react is installed in my node_modules folder, so you should be able to use.

vacas5: check out another older method but another convention for creating refs: https://5a046bf5a6188f4b8fa4938a--reactjs.netlify.com/docs/refs-and-the-dom.html

PR #875, opened 2019-10-15
fix: clean dfp strings of special characters

vacas5:
Google ads are very particular about which special characters are
allowed and which are not. Many of our place names contain special latin
characters, commas or parentheses. Clean these before slugifying and
passing to DFP as key values. This will fix an issue with two places in
the upcoming BiT campaign that should be targeted.


vacas5:
@harterc1 it looks like rizzo-next used to kebab-case the strings before sending to DFP. Do we want to do something similar in dotcom frontend?

PR #877, opened 2020-10-23
fix: remove newsletter form from footer and replace with button

katelynp43:
[Monday Pulse](https://lonelyplanet-company.monday.com/boards/807550045/pulses/807550059)

In preparation to move from Salesforce to Cordial, I am removing the newsletter form from footer and replacing it with button to newsletter landing page.

Before:
Screen Shot 2020-10-23 at 2 11 12 PM

After:
Screen Shot 2020-10-23 at 2 11 40 PM


lib/data/default.json
```diff
@@ -240,269 +240,7 @@
...
-        "Virgin Islands, U.S.",
-        "Wallis and Futuna",
-        "Western Sahara",
-        "Yemen",
-        "Zaire",
-        "Zambia",
-        "Zimbabwe"
-      ],
-      "copyright_year": "2017"
+      "copyright_year": "2020"
```
vacas5, line 243: can we use a JS date so we never have to update it again?

vacas5 (approved):

lib/data/default.json
katelynp43: Yep, I just updated.

katelynp43: Oops, pushed up my testing code. Fixed and updated.

vacas5 (approved):

PR #881, opened 2021-04-22
RM Brasil from footer

jperegoy:
[JIRA](https://redventures.atlassian.net/browse/LPWEB-138)

The linter complained about unused references on commit, both of which were untouched in this work so I forced this through with the `--no-verify` flag. If that's a no-no with this repo, let me know. Otherwise, this should merge in and then I can version and publish it which will allow me to upgrade the various repos which appear to lean on this for their footer.


vacas5 (approved):
rizzo. yeesh.

